PR #14, opened 2017-08-20
fixed printing in python

SteveDiamond:
This change is important for my unit tests. I tested it out on my machine.


bstellato:
Looks good! 

I remember we had `PySys_WriteStdout` in the past, but we changed it back to `printf` since there was no obvious reason to keep it. Happy to use `PySys_WriteStdout` if it helps.

PR #26, opened 2017-12-28
Update mpc.rst

tink3r:
The string set in res.info.status of the Python and Matlab interface seems to be lower case, not upper. Tested usind the current Windows binaries (osqp-0.2.1-matlab-windows64).


bstellato:
Thank you for spotting the typo! As you noticed, all the statuses are lowercase.

PR #28, opened 2018-01-05
Fortran interface

bstellato:
Fortran interface development and discussion. We need to:

- [ ] Change the function calls so that OSQP remembers its status (needed if called multiple times without refactoring)
- [ ] Add functions to update the solver parameters
- [ ] Add documentation


bstellato:
Moved to a separate repository as the other interfaces: https://github.com/oxfordcontrol/osqp-fortran

PR #32, opened 2018-01-06
Add time_limit option

### LINKED ISSUE #27: Add time_limit option
It is useful in soft real-time mpc to be able to have large optimisations run in the free time between mpc iterations. For this to work it must be possible to interrupt the solver at arbitrary times.

This is almost possible using check_termination with a value of 1 and the ctrl-c handler. However, this does not work on windows as signal handlers are per-process so this breaks with multiple osqp solvers running (and does not restore the current signal handler).

I'd suggest adding an interrupt callback that executes at the same time the ctrl-c handler and can be enabled irrespective of the value of the CTRLC build option using:

`c_int osqp_set_interrupt_callback(OSQPWorkspace* work, c_int (*callback)(void* opaque), void* opaque);`

The callback is disabled if a value of `NULL` is passed to `callback`.

I'm happy to implement this. Just want feedback.

ebarnard:
Closes #27


ebarnard:
@bstellato This is now ready for review.

bstellato:
That looks good. I made some changes

- I prefer to have the `time_limit` value to `0` when it is disabled so that it mimics the behavior of the other settings.
- I made sure the `time_limit` now refers to the total run time (`setup_time + solve_time`) and not just to the `solve_time`. This is controlled using the `first_run` variable.
- I have added the `time_limit` variable in the code generation phase.

ebarnard:
I'm not sure making time_limit include `setup_time` on the first run makes sense:
- Setup cannot be interrupted so will take as long as it needs irrespective of the value in `time_limit`.
- If you set `time_limit` after calling `osqp_setup` it would be very surprising for this to include `setup_time`.
- This only applies to `setup_time` and not the time for any re-factorisations (update P/A).

bstellato:
My idea is that we should limit the total run time. Many people just want to solve QPs on desktop platform and put a time limit. They do not care about the time spent in the setup or solve phases (e.g., when they call OSQP from CVXPY). That is exactly what GUROBI or other commercial solvers do.

> Setup cannot be interrupted so will take as long as it needs irrespective of the value in `time_limit`.

We can add some time limit checks in the setup function and interrupt earlier in case.

> If you set `time_limit` after calling `osqp_setup` it would be very surprising for this to include `setup_time`.

If we set a different `time_limit` with the update function, we should add a check to see if the current setup time violates that. If it is true we set the status value appropriately.

> This only applies to `setup_time` and not the time for any re-factorisations (update P/A).

I do not understand this point. If we solve the problem multiple times, we use only the latest solve time as a total run time. This value includes refactorizations if they happen in the last solve. If we refactorizations happen in the previous solutions, we ignore them exactly as we ignore the time taken to solve the previous parametric QPs.

ebarnard:
> My idea is that we should limit the total run time. Many people just want to solve QPs on desktop platform and put a time limit. They do not care about the time spent in the setup or solve phases (e.g., when they call OSQP from CVXPY). That is exactly what GUROBI or other commercial solvers do.

> We can add some time limit checks in the setup function and interrupt earlier in case.

Having `time_limit` limit total time makes sense only is setup is fully interruptible. It would be very surprising to set a 50ms total time limit and see a setup phase for a large problem take a few seconds.

> If we set a different time_limit with the update function, we should add a check to see if the current setup time violates that. If it is true we set the status value appropriately.

That means that you are setting a time limit for an operation that has already happened. A call to `update_time_limit` should only limit the time of operations that take place in the future.

> I do not understand this point. If we solve the problem multiple times, we use only the latest solve time as a total run time. This value includes refactorizations if they happen in the last solve. If we refactorizations happen in the previous solutions, we ignore them exactly as we ignore the time taken to solve the previous parametric QPs.

But refactorisations as a result of changing `P` and `A` and the initial factorisation in `osqp_setup` happen eagerly (at the time of the function call) and not during the call to `osqp_solve` and so aren't included in `solve_time` or `setup_time`, only rho changes are.

If osqp were modified to perform all factorisations (and other time consuming operations) during the call to `osqp_solve` and all these operations were fully interruptible then I would agree with having `time_limit` as a total time limit. Without using coroutines/fibres this is going to be time consuming (from looking at e.g. the LDL code) (see `libcoro`).

Otherwise I would prefer to rename this to `solve_time_limit` as having a total time limit that does not limit the time taken for some operations is confusing.

bstellato:
Apologies for the delay. I have two main concerns:

1. The `solve` function happens to perform factorizations as well. These are not fully interruptible and the issue you pointed out in 

> It would be very surprising to set a 50ms total time limit and see a setup phase for a large problem take a few seconds.

can exactly happen during the `solve` function.

2. Many solvers implementing the `time_limit` option do not separate the `setup/preprocessing` phase from the `solve phase`. I have a few examples.


In CPLEX they have the `timelimit` parameter https://goo.gl/b5BsJr which consists of
> The time limit for an optimizer applies to the sum of all its steps, such as preprocessing, crossover, and internal calls to other optimizers.
> In a sequence of calls to optimizers, the limit is not cumulative but applies to each call individually. For example, if you set a time limit of 10 seconds, and you call MIP optimization twice then there could be a total of (at most) 20 seconds of running time if each call consumes its maximum allotment.

In GUROBI they have ‘TimeLimit’ http://www.gurobi.com/documentation/7.5/refman/timelimit.html consisting of
> Limits the total time expended (in seconds). Optimization returns with a TIME_LIMIT status if the limit is exceeded

In IPOPT they do not have a special interrupt callback but they just state that the time limit is checked at certain places. See the `max_cpu_time` option: https://goo.gl/Htia3x
> Maximum number of CPU seconds. A limit on CPU seconds that Ipopt can use to solve one problem. If during the convergence check this limit is exceeded, Ipopt will terminate with a corresponding error message.

Nobody splits the `time_limit` in setup vs solve phases. That’s why I am concerned. For simplicity, code maintainability and library-free solutions I would try to stick with a compromise similar to what they did in Ipopt, i.e., setting a `time_limit` without having fully interruptible routines.

bstellato:
@ebarnard I have to ask you to sign the CLA before being able to merge.

ebarnard:
I completely agree that time_limit should limit the total time.

Given factorisations take place in `osqp_solve` the idea of limiting only solve time seems pretty pointless so I'm happy to land this PR as is.

However, the current time limit situation isn't ideal. In my mind this is what it should look like:
1. `time_limit` limits the time taken for a logical solve i.e. the total time of `osqp_setup`, any `osqp_update_*` calls, and `osqp_solve`.
2. `osqp_setup` and `osqp_update_*` queue any expensive operations to take place during the next call to `osqp_solve`. Frequent termination checks are added to `osqp_solve`.
3. (maybe eventually) `osqp_setup` is fully interruptible.

If you agree with 1 & 2 I'll have a look at how easy this would be to implement.

ebarnard:
@bstellato I think this is ready to merge.

bstellato:
Great! Merging this version for now. I like the idea of queuing the expensive operations inside `osqp_solve`. It will make it easier to track the execution time. However, we should make sure there are no extra memory allocations required in doing so (e.g., more matrices or vectors stored).

PR #51, opened 2018-02-27
initialize can_print

hongkai-dai:
I got a warning when I compile `osqp`, complaining that `can_print` might be uninitialized.

```external/osqp/include/osqp.h: In function 'osqp_solve':
external/osqp/src/osqp.c:410:5: warning: 'can_print' may be used uninitialized in this function [-Wmaybe-uninitialized]
   if(!can_print){
     ^
external/osqp/src/osqp.c:228:8: note: 'can_print' was declared here
  c_int can_print; // Boolean whether you can print

```
This was on an old version of OSQP, so the line number do not match exactly.


bstellato:
Looks good thanks!

PR #66, opened 2018-05-10
WIP: Refactor types

bstellato:
Refactor OSQP types related to sparse matrices. These will carry both copies of CSC and CSR matrices to speedup computations and parallelization.

To complete:

- [x] Write new types in both CSC and CSR
- [ ] Adapt internal functions to new types
- [ ] Adapt interface functions to the new types
- [ ] Change all the external interfaces


bstellato:
I think this PR would definitely help in that respect. It will be very easy to add functions to retrieve the sparsity pattern depending on the CSC or CSR representation.

I did not have time to work seriously on this but the current plan is to
* fix the internal representation with this PR
* implement the new ldl solver (we have a prototype working!) 
* implement the indirect solver and parallelizations

PR #69, opened 2018-06-04
fix build when profiling is off

ahoarau:


bstellato:
Thanks for this! Looks good.

PR #74, opened 2018-06-29
Enable compilation with VS 2008 for Python 2.7

mcg1969:
These simple fixes will enable this code to be compiled with Visual Studio 2008, which should simplify the building of a Python 2.7 module considerably. 

See https://github.com/oxfordcontrol/osqp-python/pull/6


mcg1969:
@bstellato I'm not sure why the travis-ci check says it is still pending; if you click through, you'll see that it is working just fine.

bstellato:
Looks good. Thanks!

PR #75, opened 2018-06-29
Check nonconvex problem

bstellato:
Added check for nonconvex problem when the residuals are too large (i.e., `2 * OSQP_INFTY`). I have added a unit test to check it.

Added a new output status: `OSQP_NON_CVX`

TODO:

- [ ] Update the status in all the interfaces.

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -27,6 +27,7 @@ extern "C" {
 # ifdef PROFILING
 #  define OSQP_TIME_LIMIT_REACHED (-6)
 # endif // ifdef PROFILING
+# define OSQP_NON_CVX (-7)           /* problem non convex */
 # define OSQP_UNSOLVED (-10) /* Unsolved. Only setup function has been called */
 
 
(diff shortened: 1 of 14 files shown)


include/constants.h
```diff
@@ -81,7 +81,7 @@ static const char *LINSYS_SOLVER_NAME[] = {
 # endif /* ifndef OSQP_NULL */
 
 # ifndef OSQP_NAN
-#  define OSQP_NAN ((c_float)0.0/0.0) // Not a Number
+#  define OSQP_NAN ((c_float)0x7fc00000UL) // Not a Number
```
bstellato: It makes sense. I found that solution on stackoverflow but apparently it is not so portable. It worked on my laptop. Did it give you issues on Windows? Unless we use C99, this might be not a portable solution see [this](https://stackoverflow.com/questions/6235847/how-to-generate-nan-infinity-and-infinity-in-ansi-c).

gbanjac: It works on Windows when I compile it with MinGW, but I get an error when compiling the Python interface.

PR #72, opened 2018-06-13
Namespaces fix

bstellato:
Various changes to fix namespace issues in C++ projects including OSQP.

Fixes #67 .

PR #90, opened 2018-09-18
Namespace interrupt functions and tic/toc

mlubin:
Avoids naming conflicts with ECOS.
See #72 and #67.

I added `static` to the unix version of `handle_ctrlc` because it's internal to ctrlc.c. I'm not confident enough to touch the windows version and also can't test that it works locally.

### DIFF
--- a/src/osqp.c
+++ b/src/osqp.c
@@ -103,7 +103,7 @@ OSQPWorkspace* osqp_setup(const OSQPData *data, OSQPSettings *settings) {
   // Start and allocate directly timer
 # ifdef PROFILING
   work->timer = c_malloc(sizeof(OSQPTimer));
-  tic(work->timer);
+  osqp_tic(work->timer);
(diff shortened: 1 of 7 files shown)


src/osqp.c
```diff
@@ -227,7 +227,7 @@ OSQPWorkspace* osqp_setup(const OSQPData *data, OSQPSettings *settings) {
   work->info->solve_time  = 0.0;              // Solve time to zero
   work->info->polish_time = 0.0;              // Polish time to zero
   work->info->run_time    = 0.0;              // Total run time to zero
-  work->info->setup_time  = toc(work->timer); // Updater timer information
+  work->info->setup_time  = osqp_toc(work->timer); // Updater timer information
```
mlubin: Check the alignment on these comments

bstellato:
Sounds good!

PR #102, opened 2018-10-29
Updated and added more badges

bstellato:
Updated badges. Added download stats badges.


bstellato:
Ooops that was a copy-paste mistake!

PR #126, opened 2019-04-03
Fix documentation for the update A and P&A functions

imciner2:
Fixes documentation errors in the function comments for the osqp_update_A and osqp_update_P_A.

### DIFF
--- a/src/osqp.c
+++ b/src/osqp.c
@@ -1073,7 +1073,7 @@ c_int osqp_update_P(OSQPWorkspace *work,
  *
  *
  *  If Ax_new_idx is OSQP_NULL, Ax_new is assumed to be as long as A->x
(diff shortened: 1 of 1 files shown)


gbanjac:
Thanks for the fix and for pointing out the duplicated documentation. This should be fixed in https://github.com/oxfordcontrol/osqp/issues/127 and was caused by function descriptions in both `osqp.h` and `osqp.c` files. Our convention is to document a function only in a header file. That is why I removed the function description from `osqp.c` in https://github.com/oxfordcontrol/osqp/issues/127.

Since your fix in this PR was done in `osqp.c` file, I think we should just close it since I applied your fix in https://github.com/oxfordcontrol/osqp/issues/127.

PR #132, opened 2019-04-24
Fixed printing disabled

### LINKED ISSUE #131: Build fails when PRINTING is not defined
Undefining `PRINTING` in `osqp_configure.h` results in the following build error: 

```
../../rt/3rdparty/osqp/src/util.c: In function ‘copy_settings’:
../../rt/3rdparty/osqp/src/util.c:227:3: error: implicit declaration of function ‘memcpy’ [-Werror=implicit-function-declaration]
   memcpy(new, settings, sizeof(OSQPSettings));
   ^
../../rt/3rdparty/osqp/src/util.c:227:3: error: incompatible implicit declaration of built-in function ‘memcpy’ [-Werror]
../../rt/3rdparty/osqp/src/util.c:227:3: note: include ‘<string.h>’ or provide a declaration of ‘memcpy’
cc1: all warnings being treated as errors
```

This happens because `<string.h>` is only included when `PRINTING` is defined. 

This isn't actually a problem for me (I just ran into it while testing out build flags), but I figured it was worth mentioning.

bstellato:
Fixes #131

### DIFF
--- a/include/auxil.h
+++ b/include/auxil.h
@@ -101,7 +101,7 @@ c_float compute_obj_val(OSQPWorkspace *work,
 
 /**
  * Check whether QP has solution
- * @param work Workspace
+ * @param info OSQPInfo
  */
 c_int has_solution(OSQPInfo *info);
 
--- a/include/auxil.h
+++ b/include/auxil.h
@@ -99,6 +99,11 @@ void update_y(OSQPWorkspace *work);
 c_float compute_obj_val(OSQPWorkspace *work,
                         c_float       *x);
 
+/**
+ * Check whether QP has solution
+ * @param work Workspace
+ */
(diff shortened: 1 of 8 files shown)


include/auxil.h
```diff
@@ -99,6 +99,11 @@ void update_y(OSQPWorkspace *work);
 c_float compute_obj_val(OSQPWorkspace *work,
                         c_float       *x);
 
+/**
+ * Check whether QP has solution
+ * @param work Workspace
```
gbanjac: The input argument should be `info`, not `work`.

gbanjac:
Looks good! Thanks for doing this. I just have one minor comment.

include/auxil.h
bstellato: Good catch!

PR #133, opened 2019-04-26
Updated tests to be C++ compliant

nikalra:
- Added casts from (void*) to desired types.
- Changed char* to const char*.
- Added function prototypes to the codegen_utils.py script.

This change is helpful for clients that need to implement OSQP integration tests from C++ unit testing frameworks.


gbanjac:
Looks good. Thanks for doing this!

PR #135, opened 2019-04-29
Added support for custom allocators

nikalra:
- Added support for custom malloc, calloc, realloc, and free functions that can be set via the new optional osqp_set_global_options API.
- Added tests to ensure that custom allocators were being called.

### DIFF
--- a/include/glob_opts.h
+++ b/include/glob_opts.h
@@ -12,6 +12,8 @@ extern "C" {
 // cmake generated compiler flags
 #include "osqp_configure.h"
 
+#include <stddef.h>
+
 /* DATA CUSTOMIZATIONS (depending on memory manager)-----------------------   */
 
 // We do not need memory allocation functions if EMBEDDED is enabled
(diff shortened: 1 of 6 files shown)


nikalra:
Addresses #136

goulart-paul:
If I understand what has been done, all of the memory management functions (calloc/malloc/free etc) are handled through function pointers declared as globals in this PR.   These globals are then assigned in a context dependent way when compiling, e.g. using the mx* allocators for mathworks in a matlab build.   There is seemingly nothing that stops these function pointers from being reassigned to something else later.

Unless there is a compelling reason to do so, I don't see why it is a good idea to allow reassignment of these functions after compilation as the default behaviour.   If OSQP is compiled as a shared library, then it is seemingly possible for some process to redeclare these functions as something unexpected or malicious, or even as just NULL.  

Would you be happy instead with a mechanism through which a user could just declare which function names should be used during compilation, e.g. using some cmake mechanism?   You could then still define the memory management functions as you like, but there would be no way to reassign them after OSQP is built.

If you really do want to allow updating of these functions, I believe that a cmake method would still work.   You would just define your own functions, and then write these functions as wrappers that make reference to global function pointers elsewhere which you are free to change as you like.

nikalra:
Yes, that would work for us as well!

If there were some way to define a custom header to be included and redefine the malloc/free/alloc calls, that would be immensely helpful to us!

goulart-paul:
OK, I will see what I can do along these lines.

Issue #136, opened 2019-05-06
Support using custom allocators with OSQP


nikalra:
OSQP potentially performs quite a few allocations and in a multi-threaded context, that results in a kernel-level lock for setting up the problem.

In those cases, we'd potentially like to use a custom allocator with OSQP that avoids these problems, enabling us to do the work of setting up problems asynchronously while waiting for other results.

nikalra:
@bstellato @gbanjac Would this be a workflow that OSQP would like to support? 

Being able to use custom allocators is extremely important for our project -- just trying to get some clarity if we should go the fork route for this functionality or if this change could be integrated into this repo?

gbanjac:
@nikalra Many thanks for the PR and sorry for the delayed response!

Yes, we would definitely like to integrate your PR. However, before just merging it to the `master` branch, we would like to make sure that other interfaces (that use this repo as a submodule) still work with this change. We were busy with other things, but will definitely have a look at this in the following days.

bstellato:
I have just made a PR proposing a solution via cmake #140

PR #140, opened 2019-05-17
Custom memory allocators

bstellato:
This PR adds the possibility to pass custom memory management functions. It can be done by specifying an additional header file where the custom functions are defined (or included) as `-DOSQSP_CUSTOM_MEMORY_H=myfile.h`

The file should have the structure
```c
# include <mycustomem.h>
# define c_malloc  mymalloc
# define c_calloc mycalloc
# define c_realloc as myrealloc
# define c_free myfree 
```

@nikalra does it solve your issue? #136. Feedbacks are welcome.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,3 +17,7 @@ indent_size = 4
 # Python
 [*.yml]
 indent_size = 2
+
+# CMake
+[CMakeLists.txt]
+indent_size = 4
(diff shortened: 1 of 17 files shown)


nikalra:
This would work great, thank you!

goulart-paul:
Added an example implementing malloc/free custom functions that count the net number of allocations.   These must be implemented in a separate file or library and linked to along with the osqp static library by user applications.

Changed the `cmake` syntax for the custom memory header to `-DOSQSP_CUSTOM_MEMORY=myfile.h`, since the previous way was likely to clash with include guards implemented around a custom header.

PR #142, opened 2019-05-20
Change interface of linsys solver functions

gbanjac:
The main change is that `linsys_solver->solve` function now computes the vector `(x_tilde,z_tilde)` instead of `(x_tilde,nu)` (in the non-polishing mode). OSQP should be agnostic to `nu` since indirect solvers do not need this vector.

I have also removed unused arguments of linsys solver functions.

### DIFF
--- a/lin_sys/direct/qdldl/qdldl_interface.h
+++ b/lin_sys/direct/qdldl/qdldl_interface.h
@@ -20,7 +20,7 @@ struct qdldl {
      * @name Functions
      * @{
      */
-    c_int (*solve)(struct qdldl * self, c_float * b, const OSQPSettings * settings);
+    c_int (*solve)(struct qdldl * self, c_float * b);
 
 #ifndef EMBEDDED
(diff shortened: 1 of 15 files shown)


lin_sys/direct/qdldl/qdldl_interface.h
```diff
@@ -45,6 +45,9 @@ struct qdldl {
     c_float *Dinv;  ///< inverse of diag matrix in LDL (as a vector)
     c_int   *P;     ///< permutation of KKT matrix for factorization
     c_float *bp;    ///< workspace memory for solves
+    c_float sigma;  ///< scalar parameter
+    c_float n;      ///< number of QP variables
+    c_float m;      ///< number of QP constraints
```
gbanjac: These were changed to `c_int` later.

bstellato:
Thanks a lot! I like the simpler interface. It makes it much easier to interface indirect solvers as well. I would just make sure codegen in Python and Matlab works before merging.

PR #150, opened 2019-05-29
Check return values of functions for dynamic memory allocation

gbanjac:
Fixes https://github.com/oxfordcontrol/osqp/issues/149


docs/examples/update-vectors.rst
```diff
@@ -154,28 +154,30 @@ C
         c_int n = 2;
         c_int m = 3;
 
-        // Problem settings
-        OSQPSettings * settings = (OSQPSettings *)c_malloc(sizeof(OSQPSettings));
+        // Exitflag
+        c_int exitflag = 0;
```
bstellato: I know this is only documentation, but assigning variables before declaring all the variables is not compatible with ANSI C and I would avoid it also in the actual C code.

src/auxil.c
```diff
@@ -1046,4 +1060,19 @@ c_int validate_settings(const OSQPSettings *settings) {
   return 0;
 }
 
+c_int memory_alloc_error(void) {
```
bstellato: Two suggestions:

1. I would combine these functions in something for general error handling to make it more consistent. An option is to create an `error.c` file and put there an unifying error function getting all error codes and have its declaration in `glop_opts.h` with error codes in `constants.h`. The function can be something like `c_int osqp_error(c_int ERROR_CODE);` with all the cases. It is more readable and maintainable IMHO.

2. We can create an enum with all error codes and a mapping between the enum errors and string constants like we did for linear system solvers:
```c
static const char *LINSYS_SOLVER_NAME[] = {
  "qdldl", "mkl pardiso"
};
```
That code could go into the `error.c` file after defining it with `extern` in `constants.h` as I did in the last commit for `LINSYS_SOLVERS`. I would avoid defining global variables in headers.

What do you think @gbanjac @goulart-paul?

docs/examples/update-vectors.rst
gbanjac: It should be allowed to assign values to variables when you declare them:
https://stackoverflow.com/questions/32786702/c89-and-variable-initialization

Compiling the code above with `gcc` with `-ansi` and `-pedantic` options generates no warnings.

src/polish.c
```diff
@@ -134,10 +134,13 @@ static c_int iterative_refinement(OSQPWorkspace *work,
...
+  c_int i, j, n, exitflag;
+  c_float *dz;
+  c_float *rhs;
+
+  exitflag = 0;
+
   if (work->settings->polish_refine_iter > 0) {
-    c_int i, j, n, exitflag = 0;
-    c_float *dz;
-    c_float *rhs;
```
bstellato: By mixing variables declarations and assignments I meant what I just changed in this commit. Here `exitflag=0` was assigned before declaring `dz` which should not be allowed in ansi C.

docs/examples/update-vectors.rst
bstellato: Ok that's good then. I remember getting such errors with similar code before but I guess it was related to assignments without declarations.

bstellato:
I have renamed a couple of things and removed the distinction between the memory allocation errors in workspace and polish (The error message should state in which function it appears). 

I have also removed a warning `ormat string is not a string literal (potentially insecure) [-Wformat-security]` at line 16 of `error.c`.

I like the changes! @gbanjac let me know what you think.  We should make sure python and matlab codegen are updated to match this PR.

gbanjac:
Looks good. Thanks for the edits! I will make PRs in python and matlab interfaces.

gbanjac:
> I have renamed a couple of things and removed the distinction between the memory allocation errors in workspace and polish (The error message should state in which function it appears).

It seems that the error message would now be
```
ERROR in osqp_error: Memory allocation.
```
Is there something analogous to ` __FUNCTION__` for the parent function name?

bstellato:
This is what I did in the `c_eprint` macro. However now we are two levels below because `osqp_error` calls that macro. I made a mistake ignoring that. I made a fix making `osqp_error` a macro calling an internal function `_osqp_error` taking the function name as argument. 

I think we should probably think about implementing a good logging system instead. An example in C99 is https://github.com/rxi/log.c.

gbanjac:
Looks good. Thanks for the fix!

I agree that a proper logging system would be very useful. We can merge this PR into `develop` and design a logging system in a separate PR. What do you think?

bstellato:
Sure! Let's merge this one

PR #159, opened 2019-06-20
WIP: Pg/vector algebra

goulart-paul:
Implements OSQPVectorf and OSQPVectori types, and replaces most 
vector operations on c_float* arrays with these new types.


examples/osqp_demo.c
```diff
@@ -1,5 +1,5 @@
 #include "osqp.h"
-
+#include <stdio.h>
```
gbanjac: What is `stdio.h` needed for?

include/osqp.h
```diff
@@ -51,11 +51,25 @@ void osqp_set_default_settings(OSQPSettings *settings);
...
-c_int osqp_setup(OSQPWorkspace** workp, const OSQPData* data, const OSQPSettings* settings);
+ c_int osqp_setup(OSQPWorkspace** workp,
+                  const csc*     P,
+                  const c_float* q,
+                  const csc*     A,
+                  const c_float* l,
+                  const c_float* u,
+                  c_int m,
+                  c_int n,
+   const OSQPSettings *settings) ;
```
gbanjac: The last argument should be aligned with other arguments.

src/auxil.c
```diff
@@ -81,55 +81,65 @@ void set_rho_vec(OSQPWorkspace *work) {
 
   work->settings->rho = c_min(c_max(work->settings->rho, RHO_MIN), RHO_MAX);
 
+  c_float* l           = OSQPVectorf_data(work->data->l);
+  c_float* u           = OSQPVectorf_data(work->data->u);
+  c_float* rho_vec     = OSQPVectorf_data(work->rho_vec);
+  c_int   *constr_type = OSQPVectori_data(work->constr_type);
+
   for (i = 0; i < work->data->m; i++) {
```
gbanjac: We can similarly define `c_int m = work->data->m` and use it in the loop condition.

src/auxil.c
```diff
@@ -779,69 +788,70 @@ c_int check_termination(OSQPWorkspace *work, c_int approximate) {
...
-  if (!data) {
-# ifdef PRINTING
-    c_eprint("Missing data");
-# endif
-    return 1;
-  }
-
-  if (!(data->P)) {
+  if (!(P)) {
 # ifdef PRINTING
```
gbanjac: `if (!P)`

src/auxil.c
```diff
@@ -779,69 +788,70 @@ c_int check_termination(OSQPWorkspace *work, c_int approximate) {
...
     c_eprint("Missing matrix P");
 # endif
     return 1;
   }
 
-  if (!(data->A)) {
+  if (!(A)) {
```
gbanjac: `if (!A)`

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
...
+  OSQPVectorf* view = c_malloc(sizeof(OSQPVectorf));
+  view->length = len;
+  view->values   = a->values + head;
+  return view;
+}
+
+OSQPVectorf* OSQPVectorf_view_free(OSQPVectorf *a){
+  c_free(a);
+  return OSQP_NULL;
+}
```
gbanjac: Why not having `void OSQPVectorf_view_free(OSQPVectorf *a)`?

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
+#include "lin_alg.h"
```
gbanjac: Shouldn't we include `vector.h`?

CMakeLists.txt
```diff
@@ -187,6 +188,8 @@ set(
     include/glob_opts.h
     include/kkt.h
     include/lin_alg.h
+    include/algebra_vector.h
+    include/algebra_matrix.h
```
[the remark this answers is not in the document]
goulart-paul: My goal is to allow the user to re-implement the internal operation of the vector/matrix functions, so these headers sort of define the API of those functions.   I was then going to put various implementations into the algebra library.   This follows (more or less) the pattern as with lin_sys.h

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
...
+
+void OSQPVectorf_from_raw(OSQPVectorf *b, const c_float *av){
+  c_int i;
+  c_float* bv = b->values;
+
+  for (i = 0; i < b->length; i++) {
+    bv[i] = av[i];
+  }
+}
+
```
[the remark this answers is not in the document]
goulart-paul: These are used to copy data from c_float* as supplied by the user via the API.   The `_copy` function makes copies between `OSQPVectorf*` for internal use.

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
...
+OSQPVectorf* OSQPVectorf_malloc(c_int len){
+
+  OSQPVectorf *b;
+  b = c_malloc(sizeof(OSQPVectorf));
+  if(b){
+    b->length = len;
+    if(b->length){
+      b->values = c_malloc(len * sizeof(c_float));
+      if(!b->values){
```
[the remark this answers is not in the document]
goulart-paul: I have changed it.

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
...
+OSQPVectori* OSQPVectori_copy_new(const OSQPVectori *a){
+
+  OSQPVectori* b = OSQPVectori_malloc(a->length);
+  OSQPVectori_copy(b,a);
+  return b;
+}
+
+/* Free a float vector*/
+OSQPVectorf* OSQPVectorf_free(OSQPVectorf *a){
+  if(a) c_free(a->values);
```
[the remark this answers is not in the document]
goulart-paul: Agree - fixed.

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
...
+  for (i = 0; i < b->length; i++) {
+    bv[i] = av[i];
+  }
+}
+
+void OSQPVectori_from_raw(OSQPVectori *b, const c_int *av){
+  c_int i;
+  c_int* bv = b->values;
+
+  for (i = 0; i < b->length; i++) {
```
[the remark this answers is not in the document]
goulart-paul: I assumed that the compiler would recognize that `b->length` does not change and would not do this.   The TI compiler is apparently crap at this sort of thing though, so perhaps we should make this sort of change everything in the base implementation.

algebra/vector.c
```diff
@@ -0,0 +1,800 @@
...
+    bv[i] = av[i];
+  }
+}
+
+void OSQPVectorf_set_scalar(OSQPVectorf *a, c_float sc){
+  c_int i;
+  for (i = 0; i < a->length; i++) {
+    a->values[i] = sc;
+  }
+  return;
```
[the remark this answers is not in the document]
goulart-paul: fixed.

examples/osqp_demo.c
goulart-paul: artefact of debugging efforts.   Removed.

include/osqp.h
goulart-paul: fixed.

src/auxil.c
goulart-paul: fixed.

src/auxil.c
goulart-paul: Not from this PR, but fixed anyway.

src/auxil.c
goulart-paul: Fixed.

algebra/vector.c
goulart-paul: fixed.

algebra/matrix.c
```diff
@@ -202,7 +202,7 @@ c_float quad_form(const csc *P, const c_float *x) {
 #ifdef PRINTING
         c_eprint("quad_form matrix is not upper triangular");
 #endif /* ifdef PRINTING */
-        return OSQP_NULL;
+        return -1.;
```
gbanjac: `OSQP_NULL` is defined as `0`, which is a valid result of `quad_form` function (e.g. when `P` is zero). I changed it to `-1` which makes it easier to detect the failure.

include/algebra_vector.h
```diff
@@ -330,7 +330,7 @@ c_float OSQPVectorf_dot_prod_signed(const OSQPVectorf *a,
                              const OSQPVectorf *b);
 
 /* check l <= u elementwise */
-c_int OSQPVectorf_ew_lt_eq(OSQPVectorf *l, OSQPVectorf* u);
+c_int OSQPVectorf_ew_leq(OSQPVectorf *l, OSQPVectorf* u);
```
gbanjac: I changed the name of this function to resemble the latex command for `<=`.

src/osqp.c
```diff
@@ -232,18 +230,18 @@ c_int osqp_setup(OSQPWorkspace** workp,
...
 
   // Allocate solution
   work->solution = c_calloc(1, sizeof(OSQPSolution));
   if (!(work->solution)) return osqp_error(OSQP_MEM_ALLOC_ERROR);
   work->solution->x = c_calloc(1, n * sizeof(c_float));
   work->solution->y = c_calloc(1, m * sizeof(c_float));
-  if (!(work->solution->x))            return osqp_error(OSQP_MEM_ALLOC_ERROR);
-  if (!(work->solution->y)) return osqp_error(OSQP_MEM_ALLOC_ERROR);
+  if (!(work->solution->x))      return osqp_error(OSQP_MEM_ALLOC_ERROR);
+  if (m && !(work->solution->y)) return osqp_error(OSQP_MEM_ALLOC_ERROR);
```
gbanjac: `mxMalloc` returns `NULL` when allocating zero memory, which was causing an error when solving an unconstrained problem (`m=0`).

goulart-paul: I thought this was what always happened, but apparently not.   Would maybe be better to explicitly set the pointer to NULL instead of allocating in the first place if m = NULL.   I think I did this in the OSQPVectorx types, but for the solution struct I guess you are working with bare arrays still.

include/algebra_vector.h
goulart-paul: Maybe this should be `OSQPVectorf_all_leq`, since it's behaviour is not returning an elementwise comparison as output in in a vector.

gbanjac: I agree. Can you please make the edit?

src/osqp.c
gbanjac: Yes, there were no problems with `OSQPVectorx`, only with bare arrays.

include/algebra_vector.h
gbanjac: Also, `OSQPVectorf_ew_bound_vec` could be renamed to something like `OSQPVectorf_clip_vec`.

goulart-paul: > I agree. Can you please make the edit?

Done.

algebra/vector.c
```diff
@@ -0,0 +1,975 @@
...
+  return out;
+}
+
+OSQPVectorf* OSQPVectorf_malloc(c_int length){
+
+  OSQPVectorf *b;
+  b = c_malloc(sizeof(OSQPVectorf));
+  if (b) {
+    b->length = length;
+    if (b->length) {
```
gbanjac: You could replace `if (b->length)` by `if (length)` to avoid dereferencing a pointer.

goulart-paul:
Closing this in favor of the larger api refactor

PR #164, opened 2019-07-10
fix for CRAN/solaris

goulart-paul:
Unsets the R_HOME environment variable manually when building OSQP for R on Solaris.

The existence of this environment variable causes a build fail on CRAN, since the R command R RHOME issues a warning as its first output, causing cmake to incorrectly assign R_ROOT_DIR.

This only appears to happen on the Solaris build, for reasons that remain shrouded in mystery.

NB: It would likely be harmless to call unset(ENV{R_HOME}) on any platform, but this is a minimal intervention.

### DIFF
--- a/configure/cmake/FindR.cmake
+++ b/configure/cmake/FindR.cmake
@@ -11,11 +11,22 @@ set(CMAKE_FIND_APPBUNDLE "NEVER")
 find_program(R_EXEC NAMES R R.exe)
 set(CMAKE_FIND_APPBUNDLE ${TEMP_CMAKE_FIND_APPBUNDLE})
 
+set(ENV{R_HOME} "blah")
 
 #---Find includes and libraries if R exists
 if(R_EXEC)
 
   set(R_FOUND TRUE)
+
+  # On CRAN/Solaris an R_HOME env variable is set and causes
+  # assignment of R_ROOT_DIR to fail since 'R RHOME'
+  # issues a warning as its first output.   Unset if needed.
+
+  # if((CMAKE_HOST_SOLARIS) AND (DEFINED ENV{R_HOME}))
+  if((TRUE) AND (DEFINED ENV{R_HOME}))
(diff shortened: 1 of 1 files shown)


configure/cmake/FindR.cmake
```diff
@@ -11,12 +11,23 @@ set(CMAKE_FIND_APPBUNDLE "NEVER")
...
 #---Find includes and libraries if R exists
 if(R_EXEC)
 
   set(R_FOUND TRUE)
 
+  # On CRAN/Solaris an R_HOME env variable is set and causes
+  # assignment of R_ROOT_DIR to fail since 'R RHOME'
+  # issues a warning as its first output.   Unset if needed.
+
+  # if((CMAKE_HOST_SOLARIS) AND (DEFINED ENV{R_HOME}))
```
bstellato: If it is no longer necessary, I would remove the comment here to cleanup

configure/cmake/FindR.cmake
```diff
@@ -11,12 +11,23 @@ set(CMAKE_FIND_APPBUNDLE "NEVER")
...
+  if((TRUE) AND (DEFINED ENV{R_HOME}))
```
bstellato: If I understood it correctly we are just checking that the `ENV{R_HOME}` is defined. This means we could remove the `(TRUE)` part from the condition

configure/cmake/FindR.cmake
```diff
@@ -11,12 +11,23 @@ set(CMAKE_FIND_APPBUNDLE "NEVER")
 find_program(R_EXEC NAMES R R.exe)
 set(CMAKE_FIND_APPBUNDLE ${TEMP_CMAKE_FIND_APPBUNDLE})
 
+set(ENV{R_HOME} "blah")
```
bstellato: Is this necessary in general or just for being able to reproduce the issue on our local machine? In the latter case, I would remove it.

PR #171, opened 2019-08-20
Make error macro C99 friendly

imciner2:
The C99 standard defines the __func__ macro for the function name, so that should be the preferred macro to use for C99 and later. This makes it a conditional define that uses __func__ if it is C99 or later, and __FUNCTION__ for the other cases (which will be compiler dependent).

This isn't actually an error when using C99 and GCC, but it triggers a warning that gets annoying.

### DIFF
--- a/include/error.h
+++ b/include/error.h
@@ -11,7 +11,13 @@ extern "C" {
 
 
 /* OSQP error macro */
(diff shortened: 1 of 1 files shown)


bstellato:
Thanks for spotting this!

PR #174, opened 2019-08-23
CMake updates for embedded code generation

imciner2:
The main portion of the PR is to clean up the CMake list creation so it can simplify the upkeep of the CMake for the code generation routines. This moves the list creation into a CMakeLists inside the src and include folders, so now in code generation frameworks the main CMakeLists should just add these directories instead of having to keep a separate list of the files (which can become out of date as changes are made).

Additionally, the first commit fixes a build error that was introduced recently in embedded mode, and the last commit adds building of the embedded code to the Travis script so that any changes that might break embedded compilation will be caught in CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,57 +159,11 @@ if (NOT MSVC)
     endif()
 endif (NOT MSVC)
 
-# Set sources
+# Set sources and includes
 # ----------------------------------------------
-set(
-    osqp_src
-    src/auxil.c
-    src/cs.c
-    src/ctrlc.c
-    src/error.c
-    src/kkt.c
-    src/lin_alg.c
-    src/lin_sys.c
-    src/osqp.c
-    src/polish.c
-    src/proj.c
-    src/scaling.c
-    src/util.c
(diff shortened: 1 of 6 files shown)


imciner2:
Ok, this is ready for review now. I have also modified the printing test so now it should actually pass (it will timeout in that computation).

Once this is merged, it would be good to update the submodules for the develop branches in the repos with code generation. Once that is done, I will make the modifications to the Matlab generation scripts to work with this.

gbanjac:
Thanks a lot for this PR! I have made a branch `cmake_fixes` so it would probably be better to merge your PR there. We can then make similar branches in Matlab and Python to make sure everything works. Then we can easily merge the branches to `develop` or `release-0.6.0`.

gbanjac:
Is this ready to be merged into `develop`? It seems that both Matlab and Python interfaces work.

imciner2:
Yes, I think this is ready to be merged  (along with the associated branches in Matlab and Python).

gbanjac:
They are merged now.

PR #176, opened 2019-08-24
Fix some memory errors

imciner2:
This set of commits fixes 3 main things:
* The error checking functions never actually returned from the function, so it was still possible to segfault on the main interface routines.
* The ```osqp_cleanup``` function referenced the settings struct without checking for NULL, even though it was checked before freeing it.
* There was the possibility of a small memory leak in the qdldl interface on an error.


gbanjac:
Looks good. Thanks a lot for fixing them!

PR #188, opened 2019-09-03
Add build exclude for code that needs profiling

imciner2:
There was an unguarded access to a settings member that was disabled during profiling, so building with profiling off would cause a compile error.

Also, this adds building with profiling off and user interrupt off to the Travis script, so hopefully these build errors can be caught in CI in the future.


gbanjac:
Thanks a lot of the fix!

PR #190, opened 2019-09-05
language specific c_eprint

goulart-paul:
Adds preprocessor logic to allow for interface-specific definitions of c_eprint.

The behaviour is unchanged except when `R_LANG` is defined.   In the case of `R_LANG`, `c_eprint` is define directly as the R native print function, avoiding usage of `__FUNCTION__ ` and `__VA_ARGS__`.   These cause difficulties with CRAN warnings on some platforms which insist on building with`-Wpedantic`, specifically the fedora CRAN test server.


goulart-paul:
Killing this and trying again.

PR #191, opened 2019-09-05
interface-specific c_eprint config

goulart-paul:
Adds preprocessor logic to allow for interface-specific definitions of `c_eprint`.

The behaviour is unchanged except when `R_LANG` is defined. In the case of `R_LANG`, `c_eprint` is define directly as the R native print function, avoiding usage of `__FUNCTION__` and `__VA_ARGS__`. These cause difficulties with CRAN warnings on some platforms which insist on building with `-Wpedantic`, specifically the fedora CRAN test server.

### DIFF
--- a/include/glob_opts.h
+++ b/include/glob_opts.h
@@ -120,6 +120,10 @@ typedef float c_float;  /* for numerical values  */
 
 # endif // end EMBEDDED
 
+# ifdef PRINTING
+#  include <stdio.h>
+#  include <string.h>
+
 /* informational print function */
 #  ifdef MATLAB
 #   define c_print mexPrintf
--- a/include/glob_opts.h
+++ b/include/glob_opts.h
@@ -120,33 +120,9 @@ typedef float c_float;  /* for numerical values  */
 
 # endif // end EMBEDDED
 
-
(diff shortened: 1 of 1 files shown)


include/glob_opts.h
```diff
@@ -155,11 +135,15 @@ typedef float c_float;  /* for numerical values  */
...
-#  endif /* ifdef MATLAB */
-
-/* Print error macro */
-#  define c_eprint(...) c_print("ERROR in %s: ", __FUNCTION__); c_print(\
-    __VA_ARGS__); c_print("\n");
+#  endif /* c_print configuration */
+
+/* error printing function */
+#  ifdef R_LANG
+#   define c_eprint Rprintf
```
bstellato: I would add a small comment on why it is defined differently than other languages (easier for us to remember later).

bstellato:
Thanks! I think it works for now. I would add a proper logger in the future though since now it is hard to see in R if the message is an error or a simple output printing. https://github.com/oxfordcontrol/osqp/issues/152

imciner2:
Yes, adding the proper logging function should be done for v1.0.

PR #193, opened 2019-09-05
update LDL etree error handling

goulart-paul:
Adds more detailed informational messages when `QDLDL_etree` returns an error code.    This is for compatibility with https://github.com/oxfordcontrol/qdldl/pull/25.

Note that I have not yet updated the qdldl submodule target.   I will do so when we tag a v.0.1.4 release for QDLDL for the above fix.


imciner2:
@goulart-paul I just updated QDLDL to 0.1.4 and the builds are passing, is there anything else to do for this?

goulart-paul:
I think there is nothing else to do for this, but we will need to update the submodule in the next osqp to this new version.

PR #198, opened 2019-09-11
Namespaces fix

goulart-paul:
Previously #72.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ docs/doxygen_out
 .ycm_extra_conf.py
 *.swo
 *.vscode/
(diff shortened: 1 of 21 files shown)


imciner2:
@mlubin @bstellato are the changes from this PR still needed?

bstellato:
I think we already fixed it for most of the API functions that could conflict with other software.

imciner2:
I'll close this PR then. If there are any other namespace issues, we can open a new PR against the more recent code.

mlubin:
I'm not aware of any current conflicts, but it's a good general practice to prefix all exported symbols consistently.

PR #201, opened 2019-09-11
WIP: cmake reconfiguration

goulart-paul:
Rewrite of the cmake build files to allow for future modularity.


imciner2:
> Getting QDLDL via cmake rather than via a submodule will break the osqp-r R interface, since the R people require that the OSQP package be buildable without cmake. On systems without cmake we use a conventional `Makefile`.

Won't this refactor break that flow anyway? I thought the plan was to reduce the amount of conditional compilation by using CMake to add the needed files to the target (so then the platform code is determined by CMake at configure time). So doing that would almost certainly cause other issues in the OSQP-R interface build process.

goulart-paul:
> > Getting QDLDL via cmake rather than via a submodule will break the osqp-r R interface, since the R people require that the OSQP package be buildable without cmake. On systems without cmake we use a conventional `Makefile`.
> 
> Won't this refactor break that flow anyway? I thought the plan was to reduce the amount of conditional compilation by using CMake to add the needed files to the target (so then the platform code is determined by CMake at configure time). So doing that would almost certainly cause other issues in the OSQP-R interface build process.

Yes, a refactor of our source files will break the `make` based build system for osqp-r that doesn't use `cmake`.    I assume that that just means that we will need to manually specify the combination of source files that should be used when building osqp for R, i.e. roughly what already happens with the R Makefile, just with a different list of sources.

Dropping QDLDL as a submodule is a different thing, because it means that the source files won't be there at all.   I suppose we could try to clone QDLDL manually from within the R makefile as a workaround.

imciner2:
> > > Getting QDLDL via cmake rather than via a submodule will break the osqp-r R interface, since the R people require that the OSQP package be buildable without cmake. On systems without cmake we use a conventional `Makefile`.
> > 
> > 
> > Won't this refactor break that flow anyway? I thought the plan was to reduce the amount of conditional compilation by using CMake to add the needed files to the target (so then the platform code is determined by CMake at configure time). So doing that would almost certainly cause other issues in the OSQP-R interface build process.
> 
> Yes, a refactor of our source files will break the `make` based build system for osqp-r that doesn't use `cmake`. I assume that that just means that we will need to manually specify the combination of source files that should be used when building osqp for R, i.e. roughly what already happens with the R Makefile, just with a different list of sources.
> 
> Dropping QDLDL as a submodule is a different thing, because it means that the source files won't be there at all. I suppose we could try to clone QDLDL manually from within the R makefile as a workaround.

Do any other build platforms forbid the use of CMake?

My main point with this is that the Makefile for this is also going to become more complex since we will have to figure out the platform and act accordingly. This almost seems like it needs a simple autoconf script to generate a makefile, but I haven't written one before.

I personally seem to have a love/hate relationship with submodules. There are cases where I like them, but there are others where I don't. In this case I am not a fan of having a submodule because in the interfaces it then means we have submodules inside submodules (which can be a pain to deal with when you need to update something or checkout the repository without a recursive clone). Perhaps @bstellato can weigh in on this.

goulart-paul:
> Do any other build platforms forbid the use of CMake?

Not that I know of.

> 
> My main point with this is that the Makefile for this is also going to become more complex since we will have to figure out the platform and act accordingly. This almost seems like it needs a simple autoconf script to generate a makefile, but I haven't written one before.

The Makefile we use for R assumes a unix-like platform and the `osqp-r` repo just comes with a pre-made `configure.h` file, i.e. we don't autogenerate anything from `configure.h.in` and just use a best guess at a reasonable configuration for R.   The R build process will use `cmake` if exists anyway, so the makefile is  mostly a mechanism for satisfying the CRAN policy of building able to build on all of their test systems.

It think it's possible to specify that cmake must exist as a system requirement in the R package, but I don't think they like that.

> 
> I personally seem to have a love/hate relationship with submodules. There are cases where I like them, but there are others where I don't. In this case I am not a fan of having a submodule because in the interfaces it then means we have submodules inside submodules (which can be a pain to deal with when you need to update something or checkout the repository without a recursive clone). Perhaps @bstellato can weigh in on this.

The R interface is a PITA, so I think we should just do whatever makes the most sense for the other interfaces.    Something can be worked out for the R interface whatever we decide, and I strongly prefer we just do what makes sense for the core code and other interfaces.

imciner2:
So, I figured out a way we can make the cmake source fetching work that should hopefully work on the R servers (I hope). I have made it so you can specify a CMake variable `QDLDL_REPO_DIR` which points to a location on the computer that has an existing copy of the qdldl repository. Then it will use that version instead of going and downloading it during the cmake configuration.

Additionally, I have made it so we can specify the qdldl versions/repos to use through the variables `QDLDL_GIT_REPO` and `QDLDL_GIT_TAG`.

imciner2:
Closing because this has effectively been merged into the `develop-1.0` branch already, and so any improvements are best made starting from there directly instead of trying to keep this branch alive.

PR #204, opened 2019-09-16
Implements an indirect linear solver based on MKL

goulart-paul:


imciner2:
There is now an MKL CG solver merged into the develop-1.0 branch based on this work. I am going to go ahead and close this PR.

PR #215, opened 2019-10-28
Verify that q, u, and l are not null

### LINKED ISSUE #214: Segmentation fault  in osqp_setup
When calling osqp_setup segment fault , but there is no any error messages.
Segmentation fault (core dumped)

imciner2:
We were never verifying that q, u, and l were actually passed into the solver before being accessed, so we would get segmentation faults. This updates the `validate_data` function to test for if these three are null.

Fixes #214

### DIFF
--- a/src/auxil.c
+++ b/src/auxil.c
@@ -815,6 +815,27 @@ c_int validate_data(const OSQPData *data) {
     return 1;
   }
 
+  if (!(data->q)) {
+# ifdef PRINTING
+    c_eprint("Missing vector q");
+# endif
+    return 1;
+  }
+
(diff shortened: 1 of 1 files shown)


bstellato:
This behavior is not correct. We should allow vectors `l` and `u` to be NULL when the problem is unconstrained. This PR forces the malloc to return a not-null pointer when passing a 0 size. However the behavior of malloc in this case is "implementation dependent". See the [manual](https://man7.org/linux/man-pages/man3/free.3.html#:~:text=DESCRIPTION%20top,successfully%20passed%20to%20free()).

PR #218, opened 2019-11-05
Make Travis deploy arm binaries automatically

bstellato:


imciner2:
Travis has been removed, so I am going to close this. We can discuss with @vineetbansal if we want to port this over to the new GitHub actions release workflow that he is developing right now.

PR #231, opened 2020-01-14
add missing define

ahoarau:


imciner2:
This has already been fixed in the development version (commit https://github.com/oxfordcontrol/osqp/commit/fdd206e6fcb12465be4d8105b5fae4650a67665c) and will be part of the next bugfix release.

PR #256, opened 2020-06-05
Fix signs in infeasibility conditions

gbanjac:
Fix https://github.com/oxfordcontrol/osqp/issues/180 and https://github.com/oxfordcontrol/osqp/issues/255

### DIFF
--- a/ci/travis/install.sh
+++ b/ci/travis/install.sh
@@ -29,11 +29,11 @@ export PATH=${DEPS_DIR}/miniconda/bin:$PATH
 hash -r
 conda config --set always_yes yes --set changeps1 no
 conda update --yes -q conda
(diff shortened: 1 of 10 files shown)


gbanjac:
This Python code is a MWE in which infeasibility is not reported without this fix:
```
import osqp
import numpy as np
import scipy.sparse as spa

P = spa.csc_matrix([0])
q = np.array([1])
A = spa.csc_matrix([[1],[1]])
l = np.array([-np.inf, 1e-4])
u = np.array([0, np.inf])

prob = osqp.OSQP()
prob.setup(P,q,A,l,u, eps_abs=1e-5, eps_rel=1e-5)
res = prob.solve()
```

bstellato:
LGTM. Thanks!

PR #260, opened 2020-06-23
Rename the global namespace literal DELTA to avoid clashes with other libraries

avikde:
I ran into issues using osqp with https://gtsam.org/ since it uses "DELTA" as a constant that was being replaced by #define DELTA in osqp code. 

I think this minor change does not affect anything in osqp, but should avoid clashes with other libraries.

If you don't mind this change, perhaps "VERBOSE" should be renamed as well since it also seems like a common name (though I did not have any particular issue with this).

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -73,7 +73,7 @@ extern const char * OSQP_ERROR_MESSAGE[];
 
 
 # ifndef EMBEDDED
-#  define DELTA (1E-6)
+#  define OSQP_DELTA (1E-6)
 #  define POLISH (0)
 #  define POLISH_REFINE_ITER (3)
 #  define VERBOSE (1)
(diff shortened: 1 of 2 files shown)


bstellato:
Thank you for pointing out this issue! I am sure it is quite common. I believe this is something we should consistently fix in the complete codebase (public API + constants). It is related to this issue: https://github.com/oxfordcontrol/osqp/issues/67

In particular, the solution @mlubin pointed out a while ago https://github.com/oxfordcontrol/osqp/issues/67#issuecomment-391152644 seems the easiest. We should probably make everything consistent.

imciner2:
This has now been fixed in the `develop-1.0` branch, where all external constants have the prefix `OSQP_` to namespace them to OSQP and prevent conflicts with other software.

PR #268, opened 2020-08-28
Update CMakeLists.txt to use correct library directory under linux.

kevinsmia1939:
I found a some files that went into /usr/lib when building on 64-bit architectures
I think that if I build for 64-bit, it should go to /usr/lib64 with this changes.


imciner2 (approved):

imciner2:
@kevinsmia1939 sorry for the delay in merging this.

PR #314, opened 2021-07-13
Github actions

vineetbansal:
All steps I found in the old shell scripts for Travis/Appveyor seem to work fine with this setup, on all 3 platforms.
I'm using a matrix of cmake flags across all platforms, but certain steps (valgrind/coveralls etc.) are only performed only on certain matrix values of the variables.

  - Build
  - Test (if `-DUNITTESTS=ON -DCOVERAGE=ON` in the build step).
  - Valgrind/Coveralls in the case when using Linux + base cmake case with `-DUNITTESTS=ON -DCOVERAGE=ON`
  - Building Sphinx Docs

For the last step, the `.html` pages are deployed in a branch called `gh-pages`. To integrate this with Github pages, someone in charge of the repo would have to go to 'Settings -> Github Pages -> Source -> select `gh-pages` as the branch, with `/` as the folder. The live site (minus the current landing pages which seem to be generated from outside the source tree), should then be available at https://osqp.github.io/osqp/, which can be linked from somewhere in the README.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,7 +107,7 @@ jobs:
           uses: imciner2/run-lcov@v1
           with:
             input_directory: '${{ runner.workspace }}/osqp/build'
-            exclude: '"$GITHUB_WORKSPACE/tests/*" "$GITHUB_WORKSPACE/lin_sys/direct/qdldl/amd/*" "$GITHUB_WORKSPACE/lin_sys/direct/qdldl/qdldl_sources/*"" "/usr/include/*"'
+            exclude: '"$GITHUB_WORKSPACE/tests/*" "$GITHUB_WORKSPACE/lin_sys/direct/qdldl/amd/*" "$GITHUB_WORKSPACE/lin_sys/direct/qdldl/qdldl_sources/*" "/usr/include/*"'
             output_file: '${{ runner.workspace }}/osqp/build/coverage.info'
           if: ${{ runner.os == 'Linux' && matrix.cmake_flags == '-DUNITTESTS=ON -DCOVERAGE=ON' && matrix.cmake_flags_extra == '' }}
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,15 +106,15 @@ jobs:
         - name: Generate coverage
           uses: imciner2/run-lcov@v1
           with:
-            input_directory: '${{ runner.workspace }}/build'
+            input_directory: '${{ runner.workspace }}/osqp/build'
             exclude: '"$GITHUB_WORKSPACE/tests/*" "$GITHUB_WORKSPACE/lin_sys/direct/qdldl/amd/*" "$GITHUB_WORKSPACE/lin_sys/direct/qdldl/qdldl_sources/*"" "/usr/include/*"'
-            output_file: '${{ runner.workspace }}/build/coverage.info'
-            if: ${{ runner.os == 'Linux' && matrix.cmake_flags == '-DUNITTESTS=ON -DCOVERAGE=ON' && matrix.cmake_flags_extra == '' }}
+            output_file: '${{ runner.workspace }}/osqp/build/coverage.info'
+          if: ${{ runner.os == 'Linux' && matrix.cmake_flags == '-DUNITTESTS=ON -DCOVERAGE=ON' && matrix.cmake_flags_extra == '' }}
 
         - name: Coveralls
           uses: coverallsapp/github-action@master
           with:
-            path-to-lcov: '${{ runner.workspace }}/build/coverage.info'
+            path-to-lcov: '${{ runner.workspace }}/osqp/build/coverage.info'
             github-token: ${{ secrets.GITHUB_TOKEN }}
           if: ${{ runner.os == 'Linux' && matrix.cmake_flags == '-DUNITTESTS=ON -DCOVERAGE=ON' && matrix.cmake_flags_extra == '' }}
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
 
         - name: Build
           run: |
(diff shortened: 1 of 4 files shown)


bstellato:
Thanks a lot @vineetbansal! This is great. 

Just a couple of comments:
- Unfortunately the actions are not running on the main repo. This is an issue @imciner2 had before when making PR https://github.com/osqp/osqp-matlab/pull/40. @vineetbansal I have added you as an admin and closed/reopened the PR but it still does not trigger GH actions checks for some reason. Maybe we need the PR to start from a OSQP branch if that's the one introducing new CI?
- It looks like this is compiling correctly on all the platforms, which is the most annoying part and this is great! Unfortunately, I do not see the test results for some reason. At the moment I am just checking the GH action [results in your fork](https://github.com/vineetbansal/osqp/runs/3059875470) but I cannot see the tests output as, for example, in a past commit [here](https://travis-ci.org/github/oxfordcontrol/osqp/jobs/753189343) (I know... the output is horribly long, and we could definitely shorten it with a better unittesting framework). Is it just hidden in the new CI or am I missing something? 
- Do you think it would be possible to push the website to the other repo with the landing page? https://github.com/osqp/osqp.org Maybe there is a nicer way instead of the workaround I added before. We can also fix this later and keep the docs url that you mentioned. An alternative, if possible, could be to use github pages for the docs with the url osqp.org/docs or docs.osqp.org but I am not sure how easy itwould be

vineetbansal:
Hi @bstellato,
- Github actions won't run on _this_ repo unless the PR is merged (because it doesn't have a `.github/workflow/*.yml` file before that happens), or is made from within this repo itself. Now that I'm admin, let me push all this stuff to a local branch on this repo and then send a PR towards master, which should trigger the CI automatically.

- Both for this PR and for the soon-to-be resubmitted one, the results of the test (if by results you mean the stdout of `osqp_tester`) can be seen by clicking on one of the 48 individual actions on the left, and provided you see a 'Test' step on the right hand side, you can click on it to expand it and see the stdout of running osqp_tester. Note that not all actions on the left have this `Test` step - some simply compile the code (i.e. they're run without `-DUNITTESTS=ON`). The name of the action will give you a hint of what's happening. If there is no value in ever _not having_ `-DUNITTESTS=ON`, then I can remove some matrix operations (the no. of actions will reduce by half to 24), and then you'll see the 'Test' for each and every action.

- Good to know about the other repo with the website code - I didn't know about it! Once this PR is deemed complete and merged, perhaps I can look into how to integrate it with the other repo? We can deploy github pages to any domain we want, but I think it would be nice to confirm that things are being rendered correctly before attempting that further integration.

.github/workflows/main.yml
```diff
@@ -0,0 +1,154 @@
...
+          if: ${{ matrix.cmake_flags == '-DUNITTESTS=ON -DCOVERAGE=ON' }}
+
+        - name: Valgrid check
+          run: |
+            sudo apt-get install valgrind
+            valgrind --suppressions=.valgrind-suppress.supp --leak-check=full --gen-suppressions=all \
+              --track-origins=yes --error-exitcode=1 build/out/osqp_tester
+          if: ${{ runner.os == 'Linux' && matrix.cmake_flags == '-DUNITTESTS=ON -DCOVERAGE=ON' && matrix.cmake_flags_extra == '-DENABLE_MKL_PARDISO=OFF' }}
+
+        - name: Generate coverage
```
imciner2, line 101: Can we use the action that we do in the QDLDL repo here (https://github.com/osqp/qdldl/blob/715d8d560e80c131b62c7b695a4ac8e52a9e2d3c/.github/workflows/ci.yml#L50)? I think it would be good to make the workflows as similar as possible so that the maintenance burden is less.

.github/workflows/main.yml
```diff
@@ -0,0 +1,154 @@
...
+        - name: Set up conda
+          uses: conda-incubator/setup-miniconda@v2
+          with:
+            auto-update-conda: true
+            python-version: ${{ matrix.python-version }}
+
+        # -----------------
+        # OS-specific setup
+        # -----------------
+        - name: Setup (Linux)
```
imciner2, line 50: I think it might be better to put the CMake generators into the matrix environment along with the other CMake stuff (and since we probably want to add a Visual Studio generator at some point to test compiling on that):
```
include:
  - os: ubuntu-latest
    cmake_generator: "Unix Makefiles"
  - os: macos-latest
    cmake_generator: "Unix Makefiles"
  - os: windows-latest
    cmake_generator: "MinGW Makefiles"
```
and then the call to CMake in the build stage would be:
```
cmake -G ${{ matrix.cmake_generator }} -S . -B build ${{ matrix.cmake_flags }} ${{ matrix.cmake_flags_extra }}
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,154 @@
+name: Main
+
+on:
+  push:
+    branches: [ master, develop, ci ]
```
imciner2, line 5:
```suggestion
    branches: [ master, develop**, ci ]
```
We have two different develop branches with the version numbers (I think this change should let it run on both of them).

.github/workflows/main.yml
```diff
@@ -0,0 +1,154 @@
...
+  pull_request:
+    branches: [ master, develop ]
```
imciner2, line 7:
```suggestion
    branches: [ master, develop** ]
```
We have two different develop branches with the version numbers (I think this change should let it run on both of them).

imciner2 (changes requested):

.github/workflows/main.yml
vineetbansal: Sure - using that action instead now.

.github/workflows/main.yml
vineetbansal: Done (needs quotes around the variable, as I discovered)

.github/workflows/main.yml
vineetbansal: Done

.github/workflows/main.yml
vineetbansal: Done

imciner2 (approved):

PR #317, opened 2021-08-07
Static website integration

vineetbansal:
With these changes, the entire site currently at https://osqp.org/ is built in the `docs/built/html` folder and deployed to the `gh-pages` branch at the root folder level, much like the existing setup in master, which only builds the sphinx documentation right now.

With the inclusion of the css/js/image assets, this is adding around 2MB to the source size. One possible optimization is to not include the `bootstrap` stuff in the source as we're doing now, but have it be included while building the site html (by putting it as a dependency in the Gemfile, for instance). My initial attempts at this were not successful (I kept getting a bootstrap dependency error), probably because the Github pages jekyll workflow is somewhat different from a bare-metal jekyll setup. If this is a desirable optimization (it will save us around 300K), then perhaps it can be made into an issue that we can revisit later?


imciner2:
Two comments:

1) We only want to run the website build on pushes to master instead of whenever we run the other checks, so I think it would be best to move the build docs workflow to its own file.

2) Once we merge #316 into master, we should merge master into this PR. Then we should exclude the new `site/` directory from the source tarball since it is the website documentation and not the source code.

bstellato:
I think it looks good now (unless I missed something!). @vineetbansal  could you please create two issues about making the website load faster and removing the bootstrap source?

bstellato (approved):

PR #318, opened 2021-08-07
Catch2

vineetbansal:
These are the minimal changes to get us started with catch2, without getting compilation errors or warnings. The resulting binary is still `osqp_tester`. We talked about a few enhancements:

- `minunit.h` can go away in the future, and `mu_assert`s replaced with a macro/function involving catch2 `CHECK`.
- Better strings representing the intention of what's tested instead of the currently placeholder function name strings for test cases and REQUIREs, like `test_constr_sparse_mat`.
- Restructuring tests to be modular using fixtures etc.

I'm not sure if we want to wait on this PR till one or more of the above are complete, but I'm creating this anyway so others can start taking a look.


imciner2:
Looks like there are some build failures on the Windows and macOS tester builds. I think it doesn't like the comparing the output of the function with the pointer in the Catch macro used there. The best way to fix that might be doing the surgery inside all the tests to just expunge the asserts and switch to that Catch macros inside the tests instead of doing the check on the function return value.

vineetbansal:
@imciner2 - The error on `macos` was due to `clang` insisting that we specify `CXX_STANDARD` for our Catch2 executable, which needs to be `11` as per Catch2 documentation. Explicitly setting this for this target:

```
    set_target_properties(osqp_tester
            PROPERTIES
                CXX_STANDARD 11
                CXX_STANDARD_REQUIRED YES
                CXX_EXTENSIONS NO
            )
```

fixes this.

For Windows, again the error was due to functions defined as `const char *` returning `0` as a return value to indicate no error. In the spirit of making minimal changes to the codebase to get things to pass before further surgery resumes, I'm returning a blank string instead of `0` in functions being tested, while still using `mu_assert`. In the next iteration (after discussing this with you guys), I'll change the `mu_assert`s to either `REQUIRE` or `CHECK`.

vineetbansal:
The way `mu_assert`s are used in the code is to check for conditions sequentially, and error out immediately (the `return` statement in the current macro definition causes the short circuiting to happen). For example:

```
  // Setup correct
  mu_assert("Primal dual infeasibility test 1: Setup error!", exitflag == 0);

  // Solve Problem
  osqp_solve(work);

  // Compare solver statuses
  mu_assert("Primal dual infeasibility test 1: Error in solver status!",
            work->info->status_val == OSQP_SOLVED);
```

I could replace the `mu_assert`s with a Catch2 `REQUIRE` (which would be closer to the original implementation here than a `CHECK`), but then we'll lose the useful messages that are specified above.

A good middle ground is to still use `mu_assert` but to redefine it as:
```
#define mu_assert(msg, pred) do { INFO(msg); REQUIRE(pred); } while((void)0, 0)
```
as per a suggestion at https://github.com/catchorg/Catch2/issues/929#issuecomment-308663820.

That's what I've done here now. So our `minunit.h` now looks like:
```
/* OSQP TESTER MODULE */

#define mu_assert(msg, pred) do { INFO(msg); REQUIRE(pred); } while((void)0, 0)
#define TESTS_TOL 1e-4 // Define tests tolerance
```

And the total number of assertions has gone up from 29 to 218.

vineetbansal:
I've made the changes we discussed in the last couple of meetings and I think this is ready to be merged, with the understanding that a deeper refactoring of the unit tests (to restructure them based on use cases, fixtures etc) will come in a later PR:

1. `minunit.h` is gone, and its functionality collapsed into `osqp_tester.h`
2. `test_catch2.cpp` is gone, and its functionality collapsed into `osqp_tester.cpp`
3. All tests now return `void` (they had initially been modified to return an `int` after our initial refactoring of getting away from `static const char*`s) and the code invoking them simply relies on their being able to run to completion without throwing errors (i.e. `REQUIRE (test_me() == 0)` replaced by `REQUIRE_NOTHROW(test_me())`

bstellato:
I think it looks good to me. I have a small question: is there a specific reason for using `REQUIRE_NOTHROW` instead of `CHECK`? It looks like `REQUIRE_NOTHROW` is used for exceptions instead of simple checks. Apart from this detail, I would be happy to finally merge this PR.

vineetbansal:
@bstellato - both `REQUIRE` and `CHECK` check for boolean conditions. The only difference is what they choose to do in case the condition is not met. Our original `mu_assert` was written to have the same behavior as Catch2's `REQUIRE`, which is what `mu_assert` now expands to.

However, to test the completion of individual test cases, we can't directly use the `REQUIRE` macro, since these test cases don't return anything at all.

A previous iteration of this PR had modified all test cases to `return 0` in case of success, but this was artificial and required littering the calling code with checks against 0. During our discussions, we decided on going back to the version where the tests return nothing, and we just check if went to completion without throwing a runtime error (or a failing `REQUIRE` within the tests themselves). `REQUIRE_NOTHROW` is the natural choice if we want to do that.

imciner2:
What can the OSQP code throw though to make that false? My understanding is the library is still being compiled in C mode when linked to the tester suite, so I am not sure we have any exceptions that could even occur, so is this just testing if Catch threw an assertion during the test and displaying that?

vineetbansal:
@imciner2 - yes, you're right. I've spent so many years with Python now that I forgot there are (almost) no runtime exceptions in compiled code. I'm not intimately familiar with OSQP's implementation, but yes - unless the code deep down throws exceptions, looks like currently this will catch any failures with any nested `mu_assert`s and display them.

imciner2:
Maybe we can remove the `REQUIRE_NOTHROW` checks then and just call the function in the testcases? I don't think they actually add value to our test suite, and at first glance someone who isn't familiar with the OSQP code might be fooled into thinking we throw assertions in our code.

vineetbansal:
@imciner2 - great suggestion! Done now.

tests/osqp_tester.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+
+TEST_CASE( "test_basic_qp2", "[multi-file:5]" ) {
+    SECTION( "test_basic_qp2_solve" ) {
+        test_basic_qp2_solve();
+    }
+#ifdef ENABLE_MKL_PARDISO
+        SECTION( "test_basic_qp2_solve_pardiso" ) {
```
imciner2, line 95:
```suggestion
    SECTION( "test_basic_qp2_solve_pardiso" ) {
```

imciner2 (approved):
Just a minor indentation point, but yes this is good to merge.

Issue #323, opened 2021-09-01
Version header file is missing


Garfield753:
When building the current master from source (cmake 3.10.2, ubuntu 18), the file `version.h` does not get installed/copied.

Commands:
~~~
cd ~
mkdir osqp_ws
cd osqp_ws
git clone --recursive https://github.com/oxfordcontrol/osqp
cd osqp
mkdir build
cd build
cmake -G "Unix Makefiles" ..
make -j
sudo make install
~~~

The file `cmake_install.cmake` which is generated in the build folder lists a bunch of headers except `version.h`:

~~~
...
if("x${CMAKE_INSTALL_COMPONENT}x" STREQUAL "xUnspecifiedx" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/include/osqp" TYPE FILE FILES
    "/home/ubuntu18/osqp_ws/osqp/include/auxil.h"
    "/home/ubuntu18/osqp_ws/osqp/include/constants.h"
    "/home/ubuntu18/osqp_ws/osqp/include/error.h"
    "/home/ubuntu18/osqp_ws/osqp/include/glob_opts.h"
    "/home/ubuntu18/osqp_ws/osqp/include/lin_alg.h"
    "/home/ubuntu18/osqp_ws/osqp/include/osqp.h"
    "/home/ubuntu18/osqp_ws/osqp/include/osqp_configure.h"
    "/home/ubuntu18/osqp_ws/osqp/include/proj.h"
    "/home/ubuntu18/osqp_ws/osqp/include/scaling.h"
    "/home/ubuntu18/osqp_ws/osqp/include/types.h"
    "/home/ubuntu18/osqp_ws/osqp/include/util.h"
    "/home/ubuntu18/osqp_ws/osqp/include/cs.h"
    "/home/ubuntu18/osqp_ws/osqp/include/polish.h"
    "/home/ubuntu18/osqp_ws/osqp/include/lin_sys.h"
    "/home/ubuntu18/osqp_ws/osqp/include/ctrlc.h"
    )
endif()
...
~~~

When adding `"/home/ubuntu18/osqp_ws/osqp/include/version.h"`  to the file by hand it installs fine. However, the stored version number does not seem to be correct (if that is critical):

~~~
/**
This file is replaced by an auto-generated version.h
with an OSQP_VERSION obtained from a variable supplied
to cmake
*/

#ifndef OSQP_VERSION
[...]

PR #327, opened 2021-09-02
Update CMakeLists.txt

vineetbansal:
Added version.h to osqp_headers as per issue #323. `make install` installs `version.h` at `<prefix>/include/osqp` folder.


imciner2 (approved):
This looks fine to me. Do you have any idea why the version.h in the issue seems to be all 0? It looks like they are cloning the repo, so it should pick up on the version that way - shouldn't it (or am I misunderstanding the new mechanism)?

PR #340, opened 2021-09-21
Port `cuda-1.0` developments to `develop-1.0`

gbanjac:
We should find a way to compile and test the code with `CUDA_SUPPORT` flag enabled.  As far as I am aware, one can use GitHub actions to compile the CUDA code, but not to test it. We should find a way to setup self-hosted runners.


gbanjac:
I compiled and successfully ran the unittests with the `CUDA_SUPPORT` flag enabled (on our lab server).

.github/workflows/main-cuda.yml
```diff
@@ -0,0 +1,65 @@
...
+  build_and_test:
+      runs-on: [self-hosted, gpu]
+
+      strategy:
+        fail-fast: false
+
+        matrix:
+          os: [ubuntu-latest]
+          python-version: [3.9]
+          cmake_flags: ['', '-DUNITTESTS=ON -DCUDA_SUPPORT=ON']
```
gbanjac, line 22: What's the purpose of the first element in the list?

PR #349, opened 2021-10-08
Various changes in the API, settings, etc

gbanjac:
I have changed quite a few things in this PR.

**Settings**
- added CG settings to the `OSQPSettings` structure
- added a setting `device` that specifies device onto which algebra functions are executed (e.g. if one has access to multiple GPUs)
- renamed the `polish` setting to `polishing` (to avoid confusion with the function `polish`)
- renamed the `warm_start` setting to `warm_starting` (for the same reason)
- changed the signature of the `validate_settings` function so that it can be used from both `osqp_setup` and `osqp_update_settings`
- removed the convention that `time_limit=0.0` means that terminating the algorithm based on time limit is disabled; instead, the default value is just a very big number which is never reached

**API**
- solver status codes are now defined in `enum osqp_status_type`
- a new function `osqp_update_settings` that replaces a number of functions for updating individual settings
- a new function `osqp_update_data_vec` that updates `q`, `l`, `u` in a single function; any of these pointers can be set to `NULL` in which case the corresponding vector won't be updated
- a new function `osqp_update_data_mat` that is similar to the previous `osqp_update_P_A`
- removed compiler flags from user-exposed structures
- prepended all constants with `OSQP_`
- changed the inequalities in infeasibility conditions to be `< 0` rather than `< eps`
- renamed `pri_res` and `dua_res` to `prim_res` and `dual_res`


gbanjac:
- Since the choice of a linear system solver is very much related to the algebra used, I was even thinking of moving linear system solvers in the `algebra` folder. For instance, the CUDA PCG can be moved to `algebra/cuda/linsys/indirect`. What do you think?
- Regarding the infeasibility check, we were discussing between `< -eps` (a more restrictive condition, which was there earlier) and `< eps` (a relaxed condition, which is there atm). I have changed the check to `< 0`, so that we neither relax nor restrict that condition.

gbanjac:
I agree. I will not make any changes to the folder structure in this PR. Can do it once MKL is added.

gbanjac:
I think the PR is ready to be merged. @imciner2, @bstellato What do you think?

bstellato:
I double checked again and it LGTM. I would make sure to remember to fix the interfaces accordingly. Most importantly, with constants `OSQP_` and `prim_res` `dual_res`. Thanks again for this great PR @gbanjac

PR #359, opened 2021-12-01
Adding osqp include directories for clients to interface with osqpstatic

vineetbansal:
I was trying to compile some client code against `osqp::osqpstatic` and noticed that in all the refactoring of `CMakeLists.txt`, I had removed the include_directories for linking with `osqp::osqpstatic` and forgot to put it back. This change will bring this back in line with how `develop-1.0` behaved in this regard before PR #352 was merged, as well as the current behavior on `master`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,7 @@ endif()
 # Rename compile-time constants & configure
 # ----------------------------------------------
 # If we are creating any OSQP_* compile-time constants from CMake variables, do so here.
-set (OSQP_ALGEBRA ${ALGEBRA})
+set(OSQP_ALGEBRA ${ALGEBRA})
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/configure/osqp_configure.h.in
                ${CMAKE_CURRENT_SOURCE_DIR}/include/osqp_configure.h NEWLINE_STYLE LF)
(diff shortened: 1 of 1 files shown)


gbanjac (approved):

CMakeLists.txt
```diff
@@ -273,6 +273,9 @@ function(OSQP_BUILD_STATIC_LIB)
     # Transitive dependency on OBJECT library does not work. See https://gitlab.kitware.com/cmake/cmake/-/issues/18682
     target_sources(osqpstatic PRIVATE $<TARGET_OBJECTS:qdldlobject>)
   endif()
+  target_include_directories(
+    osqpstatic PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+                      "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${CMAKE_INSTALL_INCLUDEDIR}/osqp>")
```
imciner2, line 278: I don't think this variable will be properly populated until line 333 because this variable is included inside the `GNUInstallDirs` CMake file that we include there. We probably need to move that include to the beginning of the file so that we can reference the path properly.

imciner2 (changes requested):

CMakeLists.txt
vineetbansal: Thanks @imciner2 - Fixed now.

PR #368, opened 2021-12-10
Clean up internal header linkage statements

gbanjac:
It turns out we don't need `extern "C"` only in our public API headers, but also at the interface between our main code and the algebra since the latter can be implemented in C++.


imciner2 (approved):
Looks good to me.

PR #373, opened 2021-12-14
Code generation in C

gbanjac:
Added codegen functionality in C.

TODO:
- Generate tailored CMake file for embedded applications
- Add a codegen test


src/codegen.c
```diff
@@ -0,0 +1,591 @@
...
+/**********
+* Settings
+***********/
+
+static void write_settings(FILE               *f,
+                           const OSQPSettings *settings,
+                           const char         *prefix){
+  
+  fprintf(f, "/* Define the settings structure */\n");
+  fprintf(f, "OSQPSettings %ssettings = {\n", prefix);
```
imciner2, line 122: I think it might be cleaner/clearer if there is an underscore between the prefix and the name of the variable (i.e. `%s_settings`).

src/codegen.c
```diff
@@ -0,0 +1,591 @@
...
+* Info
+*******/
+
+static void write_info(FILE           *f,
+                       const OSQPInfo *info,
+                       const char     *prefix){
+  
+  fprintf(f, "/* Define the info structure */\n");
+  fprintf(f, "OSQPInfo %sinfo = {\n", prefix);
+  fprintf(f, "  \"%s\",\n", info->status);
```
imciner2, line 164: Does it make sense to populate these with non-default values as its starting values? This is the solver return info, so it would seem better to start it out as being default values (e.g. use `OSQP_UNSOLVED`).

src/codegen.c
```diff
@@ -0,0 +1,591 @@
...
+***********/
+
+static void write_solution(FILE       *f,
+                           c_int       n,
+                           c_int       m,
+                           const char *prefix){
+
+  fprintf(f, "/* Define the solution structure */\n");
+  fprintf(f, "c_float %ssol_x[%d];\n", prefix, n);
+  if (m > 0) fprintf(f, "c_float %ssol_y[%d];\n", prefix, m);
```
imciner2, line 193: I think an underscore should also follow the prefix on these internal variables as well.

imciner2 (commented):
Some preliminary comments on this draft. My largest comments are about writing out the `c_int` types to the file (i.e. if only `%d` will work or if we need `%lld` for the long long case, if it is allowed in embedded), and about adding an underscore between the prefix and variable names.

CMakeLists.txt
```diff
@@ -124,6 +124,12 @@ if(OSQP_CUSTOM_MEMORY)
   message(STATUS "User custom memory management header: ${OSQP_CUSTOM_MEMORY}")
 endif()
 
+if(NOT DEFINED EMBEDDED
+   AND ALGEBRA_DEFAULT)
+  option(OSQP_CODEGEN "Enable code generation" ON)
```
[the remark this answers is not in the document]
imciner2: I have gone ahead and made this change and pushed it.

src/codegen.c
```diff
@@ -0,0 +1,591 @@
...
+static void write_veci(FILE        *f,
+                       const c_int *veci,
+                       c_int        n,
+                       const char  *name){
+
+  c_int i;
+  if (n) {
+    fprintf(f, "c_int %s[%d] = {\n", name, n);
+    for (i = 0; i < n; i++) {
+      fprintf(f, "  %i,\n", veci[i]);
```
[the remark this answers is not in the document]
imciner2: Since we want to have C89 compliant code, I have disabled the use of `long long`, so this isn't an issue.

src/codegen.c
imciner2: After discussion, it will be expected the user provide an underscore in their prefix if they want to have underscores separating the prefix from the variable name.

src/codegen.c
imciner2: After discussion, it will be expected the user provide an underscore in their prefix if they want to have underscores separating the prefix from the variable name.

include/osqp_api_constants.h
```diff
@@ -43,7 +43,8 @@ enum osqp_error_type {
     OSQP_NONCVX_ERROR,
     OSQP_MEM_ALLOC_ERROR,
     OSQP_WORKSPACE_NOT_INIT_ERROR,
-    OSQP_ALGEBRA_LOAD_ERROR
+    OSQP_ALGEBRA_LOAD_ERROR,
+    OSQP_FOPEN_ERROR
```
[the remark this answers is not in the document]
imciner2: I have now added a string to the error array that corresponds with this error code.

imciner2:
Ok, I believe this code generation system is now basically complete, with the exception of how we initialize the info struct.

The way this now works is that the C code will generate three files:
* `prefixworkspace.h` exporting the solver
* `prefixworkspace.c` defining the solver and data
* `osqp_configure.h` defining the settings for the embedded OSQP code

Then the CMake build step will setup a flater directory structure in the build directory underneath the `codegen_src` directory that contains all the source and header files needed for the embedded implementation. The user can then just copy those files and the three generated files into their project, add them to their own build system, and have a working OSQP solver. There is now no more need for CMake to be run to actually build the OSQP embedded code into a library to then use.

More than one solver workspace can be included into an embedded project, however only one set of configuration options for OSQP can be used (e.g. all solvers must use either floats or doubles, there isn't the choice to mix them).

Right now, the configure settings passed into the code generation contain several options that aren't actually supported, but I have a work-in-progress branch to move the remaining parts over to an implementation that allows for custom functions (e.g. the timing and interrupt/ctrl-c support), which will allow them to be used in embedded code.

I have also forced the embedded code to run with `DLONG=OFF` to disable `long long` types (since they are not in the C89 standard, and we want to be C89 compliant).

As an additional part, I added a new API function to get the error message string for a given error code, so in an embedded application users could still get a description of errors if they wanted to save them somewhere (e.g. a system log file or similar place).

imciner2:
This is now ready. I would like to merge it soon.

bstellato (changes requested):
Per our discussion `mode1.c` -> `embedded_mode1.c` (same for 2). Can you please create two issues for:
1. C++ memory management for rest of testing
2. Update docs for codegen

bstellato (approved):
Brilliant work! Thanks so much @gbanjac and @imciner2 for this! LGTM now!

imciner2:
Its merged :tada: :fireworks:.

Issue #143, opened 2019-05-20
Implement a more efficient update of the KKT matrix


gbanjac:
When a small number of entries in `P` or `A` changes, it is possible to pass only indices of entries that need to be updated. However, the entire `P` and `A` matrices are then sent to linsys solver `update` function to update the KKT matrix.

This can be implemented more efficiently.

goulart-paul:
Sort of.   Look [here](https://github.com/osqp/osqp/blob/79b4f8614548b9df95ef136efa5363d2f5fc6d50/src/osqp_api.c#L1016).

Whether or not a partial matrix update is triggered depends on the scaling settings.   The problem is that you can't directly repopulate the KKT system with partially revised (A,P) data because the user data is not scaled, but the KKT data is.    You could scale as you passed it through from the user, but the scaling could be poor at that point so we (I think) just do the scaling all over again.

Barring some indication that the full KKT data overwrite is a bottleneck for someone I think it's best to close this issue and leave the implementation as it stands.

PR #374, opened 2021-12-15
Rewrite of the KKT assembly process

goulart-paul:
A rewrite of the KKT matrix construction process to avoid the use of an intermediate matrix in triplet format.    These functions are now allocation free except for the space actually required for the KKT matrix itself.   Code is also restructured to avoid huge functions.

I have implemented support for both CSC and CSR format KKT matrices and tested against the MKL and QDLDL algebra libraries.   

The interface to the linear solvers has been changed since some of the arguments were no longer needed or could be consolidated.

@gbanjac : I have no idea what will happen if you try this against the CUDA algebra since I haven't tried it, but it will probably fail.   I think required changes would be minimal though.   Let's see what people think of the implementation for QDLDL/MKL before looking at it.


algebra/_common/kkt.c
```diff
@@ -201,44 +335,37 @@ csc* form_KKT(c_float*    P_x,
...
+                  c_float     param1,
+                  c_int       format) {
 
-  c_int i, j; // Iterations
+  c_int j; // Iterations
 
   // Update elements of KKT using P
-  for (i = 0; i < P_p[n]; i++) {
-    KKT->x[PtoKKT[i]] = P_x[i];
+  for (j = 0; j < P->p[P->n]; j++) {
```
gbanjac, line 347: You may want to define `c_int nnzP = P->p[P->n];` and use `nnzP` in the loop condition instead.

algebra/_common/kkt.c
```diff
@@ -201,44 +335,37 @@ csc* form_KKT(c_float*    P_x,
...
+                  csc*        A,
                   c_int*      AtoKKT) {
 
-  c_int i; // Iterations
+  c_int j;
 
   // Update elements of KKT using A
-  for (i = 0; i < A_p[n]; i++) {
-    KKT->x[AtoKKT[i]] = A_x[i];
+  for (j = 0; j < A->p[A->n]; j++) {
```
gbanjac, line 363: Same comment as above.

algebra/default/lin_sys/direct/qdldl_interface.c
```diff
@@ -455,18 +441,10 @@ c_int update_linsys_solver_matrices_qdldl(qdldl_solver     *s,
                                           const OSQPMatrix *A) {
 
     // Update KKT matrix with new P
-    update_KKT_P(s->KKT,
-                 P->csc->x,
-                 P->csc->p,
-                 s->n,
-                 s->PtoKKT, s->sigma, s->Pdiag_idx, s->Pdiag_n);
+    update_KKT_P(s->KKT, P->csc, s->PtoKKT, s->sigma, 0);
```
gbanjac, line 444: It would be great if you could also address https://github.com/osqp/osqp/issues/143 in this PR. In particular, if we want to update a small number of entries in a matrix, then we can do it more efficiently than updating the whole matrix; see https://github.com/osqp/osqp/blob/develop-1.0/src/osqp_api.c#L994

gbanjac (commented):
The PR looks great! Thanks for doing this, Paul! Here are some additional comments:
- If we are not using the triplet format any more, do we even need the `nz` entry in the `csc` structure any more?
- We should be good with the CUDA algebra as it is not relying on the `algebra/_common` directory. We could, however, replicate the same logic there.

algebra/default/lin_sys/direct/qdldl_interface.c
imciner2: Are you talking about only recomputing the parts of the elimination tree after the updated values? If so, doesn't that also needs some changes in QDLDL to implement it?

gbanjac: No, I am talking about how we are updating the KKT matrix. We are exploiting the fact that we are updating only a subset of values when updating `P` and `A`, but then we are passing the whole updated matrices; see the link above.

goulart-paul:
> The PR looks great! Thanks for doing this, Paul! Here are some additional comments:
> 
> * If we are not using the triplet format any more, do we even need the `nz` entry in the `csc` structure any more?
> * We should be good with the CUDA algebra as it is not relying on the `algebra/_common` directory. We could, however, replicate the same logic there.

I think we do not, and there is even some danger in keeping since it is duplicating the value in the (n+1)^th entry of the `colptr` field.

algebra/_common/kkt.c
```diff
@@ -1,197 +1,331 @@
...
-  c_int  ptr, i, j;       // Counters for elements (i,j) and index pointer
-  c_int  zKKT = 0;        // Counter for total number of elements in P and in
-                          // KKT
-  c_int *KKT_TtoC;        // Pointer to vector mapping from KKT in triplet form
-                          // to CSC
+//add an offset to every term in the upper nxn block.
+//assumes triu CSC or CSR format, with fully populated diagonal.
+//format = 0 / CSC:  diagonal terms are last in every column.
+//format = 1 / CSR:  diagonal terms are first in every row.
+void _kkt_shifts_param1(csc* KKT, c_float param1, c_int n, c_int format){
```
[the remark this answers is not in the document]
goulart-paul: Oops.   Good idea.

algebra/default/lin_sys/direct/qdldl_interface.c
goulart-paul: Yes, I can do that as part of this PR and will add it when I fix the other stuff you suggested.    Note that for the `P` matrix update this might actually be worse than just overwriting though.   The reason is that for a partial update I think it is necessary to test whether each entry is on the diagonal, since in that case one must reapply the `P_ii + \sigma` shift to it.   It's a little bit expensive to identify whether that`s the case though, since you have to first get the row index for the entry and then look into the corresponding column to see whether it is both nonemptry and that the value you are updating is between the first and last value in the colptr range.   The check can be made a little faster since the matrix is in TRIU form and in CSC (CSR format), since in that case you can just check against the last ( first) entry in the column, but still there are 2-3 logical operations required for each entry to be updated.

Given the above, there is probably some situation in which a partial (but extensive) update could be slower than a total overwrite.   

Maybe we should have different behaviour between the 'full replacement' and 'partial replacement' cases, switching between based on the length of the update supplied.   If the length is just `nnz(P)`, then there is no need for any of the above since the whole diagonal can just be shifted at the end.

gbanjac:
> > The PR looks great! Thanks for doing this, Paul! Here are some additional comments:
> > 
> > * If we are not using the triplet format any more, do we even need the `nz` entry in the `csc` structure any more?
> > * We should be good with the CUDA algebra as it is not relying on the `algebra/_common` directory. We could, however, replicate the same logic there.
> 
> I think we do not, and there is even some danger in keeping since it is duplicating the value in the (n+1)^th entry of the `colptr` field.

I actually meant the `nz` field (not `nzmax`), which is always set to `-1` when the CSC format is used (see [here](https://github.com/osqp/osqp/blob/develop-1.0/include/csc_type.h#L19)). But you are right, we probably don't even need `nzmax`.

goulart-paul:
@gbanjac CUDA test is failing here and I don't understand why.

gbanjac:
I think it is an issue with our self-hosted runner. I've re-run [this job](https://github.com/osqp/osqp/runs/4640608055?check_suite_focus=true) that ran successfully previously, but now it fails.

@vineetbansal Could you please take a look at it?

vineetbansal:
@goulart-paul, @gbanjac - I rebooted the runner machine and re-ran the action. Seems to work fine now. Not sure what happened though!

PR #377, opened 2021-12-21
partial data matrix updates

goulart-paul:
Implements support for partial updates to P and A matrices.

NB: untested because I think unit tests will require a modification to the python testing scripts to support export of c_int arrays that are required for indicating which elements to update.

Edit : I think it would also be a good idea to write one or more extra unit tests so that we can test 1) partial update of A and/or P and 2) full update of A and/or P.  This matters for P in particular because it is implemented two different ways.

### DIFF
--- a/algebra/_common/kkt.c
+++ b/algebra/_common/kkt.c
@@ -335,36 +335,75 @@ csc* form_KKT(csc*        P,
 
 #if EMBEDDED != 1
 
-void update_KKT_P(csc*        KKT,
-                  csc*        P,
-                  c_int*      PtoKKT,
-                  c_float     param1,
-                  c_int       format) {
+void update_KKT_P(csc*         KKT,
+                  csc*         P,
+                  const c_int* Px_new_idx,
+                  c_int        P_new_n,
+                  c_int*       PtoKKT,
+                  c_float      param1,
+                  c_int        format) {
 
-  c_int j, nnzP;
-
-  // Update elements of KKT using P
+  c_int j, nnzP, Pidx, Kidx, row, offset;
   nnzP = P->p[P->n];
-  for (j = 0; j < nnzP; j++) {
-    KKT->x[PtoKKT[j]] = P->x[j];
+
+  if(Px_new_idx == OSQP_NULL || P_new_n <= 0){return;}
+
+  //if nnzP is the same as the update length update,
+  //overwrite everything and apply sigma shifts
+  //as a final step
+  if(P_new_n == nnzP)
+  {
+    for (j = 0; j < P_new_n; j++) {
+      Pidx = Px_new_idx[j];
+      Kidx = PtoKKT[Pidx];
+      KKT->x[Kidx] = P->x[Pidx];
+    }
+
+    // Update diagonal elements of KKT by adding sigma
+    _kkt_shifts_param1(KKT, param1, P->n, format);
   }
 
-  // Update diagonal elements of KKT by adding sigma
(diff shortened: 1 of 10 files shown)


algebra/_common/kkt.c
```diff
@@ -335,36 +335,75 @@ csc* form_KKT(csc*        P,
...
+void update_KKT_A(csc*         KKT,
+                  csc*         A,
+                  const c_int* Ax_new_idx,
+                  c_int        A_new_n,
+                  c_int*       AtoKKT) {
 
-  c_int j, nnzA;
+  c_int j, nnzA, Aidx, Kidx;
+
+  if(Ax_new_idx == OSQP_NULL || A_new_n <= 0){return;}
```
gbanjac, line 397: The value of `A_new_n` should be validated in `osqp_update_data_mat()`. Also, it would be better to return something like `OSQP_DATA_VALIDATION_ERROR` than `2` as it is done now (see [here](https://github.com/osqp/osqp/blob/fa222dd24c56fad4db0ed866cb5a7332fbc93b5b/src/osqp_api.c#L982)). The same fix should be made when updating the values of `P`.

goulart-paul: My thought there was that I didn't want to make separate update functions for the linear system solvers to implement, because at the moment there is just a single `update_matrices` function pointer to supply.   That means that that function should be able to handle a null pointer for either of `A` or `P` if only one is to be updated.   That's different from the update of the internal copies, since they are done separately [here](https://github.com/osqp/osqp/blob/fa222dd24c56fad4db0ed866cb5a7332fbc93b5b/src/osqp_api.c#L988).

Doesn't it still make sense to bail if a zero-length update is supplied?   I don't think what I have done there is really an error checking mechanism, since the user could legitimately be saying "here's a pointer to a an array, don't update any values from it".    I guess it could be `A_new_n == 0` to exit, but it doesn't seem to matter much.

I agree about the error code in `osqp_update_data_mat` though - I can fix that.

gbanjac: You are right. Maybe we can just include an additional check `if (A_new_n < 0)` in `osqp_update_data_mat`.

goulart-paul: After failing unit tests for ages, I realised that partial KKT updates won't work given the way that we handle scaling.   The reason relates to how we unscale and then rescale the internal data copies [here](https://github.com/osqp/osqp/blob/9dd35759fbe7d3ab3d58c54ec5cee285c110fedc/src/osqp_api.c#L996).   Note that rescaling actually computes a new equilibration, rather than just unscaling / rescaling with the factors already computed in some prior equilibration.

This means that any partial update subsequently applied to the KKT matrix will have values with different scalings applied, thereby producing incorrect results.   For now, I have forced a complete update of the KKT matrix (still allowiing partial updates to data matrices) when scaling is enabled, and allow partial updates of both P/A and the KKT data when scaling is disabled (see [here](https://github.com/osqp/osqp/blob/9dd35759fbe7d3ab3d58c54ec5cee285c110fedc/src/osqp_api.c#L1009).

I think that is the best that we can do unless we add a further option to reuse an existing equilibration, but that seems like a pain.

This is now passing unit tests, but the unit tests now cover all of the possible paths anymore.   We probably should be checking:

- partial updates of P / A (requires an index vector) 
- full updates (with no index vector)
- full updates that specify an index anyway
- updates of P and or A in isolation (with or without indices supplied, specifying 0 length updates)
- partial / full KKT updates with / without scaling

PR #386, opened 2022-01-31
Check interia of KKT matrix in Pardiso factorization to verify convexity

imciner2:
The Pardiso interface provides the inertia of the factorized matrix inside the `iparm` array as a returned value after the numerical factorization. We can use that inertia to then determine if the KKT matrix is non-convex by just checking the number of positive eigenvalues against the size of the P matrix.

### DIFF
--- a/algebra/mkl/lin_sys/direct/pardiso_interface.c
+++ b/algebra/mkl/lin_sys/direct/pardiso_interface.c
@@ -235,6 +235,14 @@ c_int init_linsys_solver_pardiso(pardiso_solver    **sp,
     *sp = OSQP_NULL;
     return OSQP_LINSYS_SOLVER_INIT_ERROR;
   }
+  if ( s->iparm[21] < n ) {
+    // Error: Number of positive eigenvalues of KKT should be the same as dimension of P
+#ifdef PRINTING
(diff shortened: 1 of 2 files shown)


imciner2:
@vineetbansal do you know what would be causing the cmake errors with MKL on the Windows CI actions here? My changes here didn't touch any of that CMake code, so there appears to be something else going wrong.

vineetbansal:
@imciner2 - looks like there's a regression w.r.t usable MKL libraries (only on Windows) on [mkl-devel 2022.*](https://anaconda.org/intel/mkl-devel) on the intel channel. For now I've pinned the version to be `mkl<2022` to get around this. Let me open up an issue on this so I can keep a close eye on this (hopefully things will stabilize on the channel soon).

All 3 commits on top of your work have been cherry-picked from the [im/apicleanup](https://github.com/osqp/osqp/tree/im/apicleanup) branch for PR #388, so let me leave comments on that PR instead.

bstellato:
Thank you, @imciner2! This is great and a very neat solution. The changes to `pardiso_interface.c` look great.

bstellato (approved):

Issue #106, opened 2018-11-18
Please respect the cmake variables BUILD_STATIC_LIBS and BUILD_SHARED_LIBS


yurivict:
Some users only want shared libs, and some - only static libs. For example, FreeBSD ports generally don't install static libs. Please honor BUILD_STATIC_LIBS and BUILD_SHARED_LIBS.

Please do this in both this and the bundled project.

Thank you.

imciner2:
Support has been added to the `develop-1.0` branch for three new build options:
* `OSQP_BUILD_SHARED_LIB` to enable/disable the building of the shared library
* `OSQP_BUILD_STATIC_LIB` to enable/disable the building of the static library
* `OSQP_BUILD_DEMO_EXE` to enable/disable the building of the demo executable (requires the static library to be built)

These will be included in the future 1.0 release.

PR #388, opened 2022-02-04
Cleanup various parts of the Cmake and file structure

### LINKED ISSUE #369: Move custom memory include to private configure header
Currently, we are using CMake to generate an include statement for the header that provides the user-specified custom memory functions, and this is being put into the main `osqp_configure.h` header file. This file is included in the public API for OSQP though, so we are effectively leaking all the custom memory logic into the user program (such as the defines `c_malloc`, `c_free`, etc.). We should split this into a separate private header that CMake will configure so that we hide these details from the user application and not leak the defines. This private header can then be used more in the future for custom timing and printing as well (so it isn't just the memory that will need it).

imciner2:
The goal of this is to cleanup the CMake and some various parts of the file structure. The main changes are:

* Add explicit flags for building the shared/static/demo executable to allow for only compiling certain ones (similar to how QDLDL is structured, and fixes https://github.com/osqp/osqp/issues/106)
* Rearrange the include directory to separate the public and private headers to make it obvious what is public API
* Define the visibility for the public headers so the functions are exposed properly on Windows (replaces the hammer used in https://github.com/osqp/osqp/pull/310)
* Remove the custom memory management header from the public header (it is an implementation detail), Fixes #369 
* Add a small CMake script to test the actual compilation of a program with a custom memory header


.github/workflows/main-cuda.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         matrix:
           os: [ubuntu-latest]
           python-version: [3.9]
-          cmake_flags: ['-DBUILD_TESTING=ON -DALGEBRA=cuda']
+          cmake_flags: ['-DBUILD_TESTING=ON -DALGEBRA=cuda -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_CUDA_HOST_COMPILER=g++-10']
```
vineetbansal, line 20: Opened issue #391 to keep track of this.

.github/workflows/main.yml
```diff
@@ -81,7 +81,7 @@ jobs:
 
         - name: Install python dependencies
           run: |
-            conda install -c intel mkl-devel
+            conda install -c intel "mkl-devel<2022"
```
vineetbansal, line 84: Opened issue #390 to keep track of this.

CMakeLists.txt
```diff
@@ -269,37 +287,41 @@ if(R_LANG)
...
-      ON
-      PARENT_SCOPE)
-endfunction()
+endif()
 
 # ----------------------------------------------
 # Shared Library - osqp
 # ----------------------------------------------
-function(OSQP_BUILD_SHARED_LIB)
-  add_library(osqp SHARED $<TARGET_OBJECTS:OSQPLIB>)
```
vineetbansal, line 298: Is there an advantage to adding the *sources* to the shared library (which also means we have to target-include-directories `osqplib_includes` to make it work) instead of the compiled objects from OSQPLIB as we were doing before? I'm sure there's a good reason, just curious for my own understanding :)

imciner2: This comes from marking the public API with `OSQP_API` to get the default visibility from the shared library. When the object files were included here, all our public functions were still with hidden visibility even though they had the `OSQP_API` annotation. The only way to fix that was to make the shared library compile the object files again (this time with the correct build flag to make the public API visible). Technically, we only need to compile the functions that have their definitions inside the public headers again to get the right visibility information in the library, but I found it simplest to just compile everything instead of having to sort out which source files had public/private functions (we probably only need to recompile about 7 functions instead of the entire codebase).

vineetbansal: Thanks @imciner2 - makes sense. I see the `BUILDING_OSQP` definition and how you're using it here. Very instructive!

CMakeLists.txt
```diff
@@ -312,41 +334,35 @@ function(OSQP_BUILD_SHARED_LIB)
...
-  set(OSQP_SHARED_LIB
-      ON
-      PARENT_SCOPE)
-endfunction()
+  # Declare that we are building the shared library to get proper symbol exports.
+  # Shared library consumers should also define OSQP_SHARED_LIB to get the library
+  # exports properly, so we do it for them in the CMake interface by defining it as
+  # a PUBLIC compile definition.
+  target_compile_definitions(osqp PRIVATE BUILDING_OSQP)
+  target_compile_definitions(osqp PUBLIC  OSQP_SHARED_LIB)
```
bstellato, line 342: This is quite interesting. Shall we document this in the docs?

bstellato (commented):
This is amazing, @imciner2! I just have a minor comment on whether we should add some information in the docs about how to deal with the API (especially in the public/private API).

PR #394, opened 2022-02-19
Rename error.h to osqp_error.h to avoid OS collisions

AustinSchuh:
On linux, error.h exists.  Depending on how your toolchain and compiler
is configured, the compiler can pick the system error.h instead.
This results in some pretty esoteric errors which are hard to debug.
Instead, prefix this file with osqp_ to avoid the collision.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -4,7 +4,7 @@ set(
     "${CMAKE_CURRENT_SOURCE_DIR}/version.h"
     "${CMAKE_CURRENT_SOURCE_DIR}/auxil.h"
     "${CMAKE_CURRENT_SOURCE_DIR}/constants.h"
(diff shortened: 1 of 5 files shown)


imciner2:
Thank you for proposing this. Please note that our next planned release is being developed on the `develop-1.0` branch, not the master branch, so please retarget the PR against that branch.

I am also somewhat confused what you mean by the compiler getting confused between the OSQP `error.h` and the system `error.h`. The header is specified as a local/non-system header in the include statements (through the use of quotes), and the include path where it is located should be placed by CMake earlier in the compilation command than the default system path (so it should take precedence). Can you be more precise about the system/compilers that are confusing the two headers please?

mhubii:
It is interesting to mention that these errors would be resolved by #401, hence I'll leave the reference here

PR #395, opened 2022-02-22
Prefix linear solver constants with OSQP to prevent clashes

imciner2:
This adds an `OSQP_` prefix to the constants/types/variables for identifying the linear solver to namespace them into OSQP and prevent possible conflicts with other software that might have similar defines.

I also shuffled around the order of the enum to make the unknown solver position 0, since previously it was at 99, and the status printing would try to index into an array with that value (so it would index into position 99 of a 2 element array). Putting it at position 0 allows for future expansion still, and puts it into a nice indexable position.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,9 @@ jobs:
         fail-fast: false
 
         matrix:
-          os: [ubuntu-latest, macos-latest, windows-latest]
+          # Specify the exact Windows image because the CMake generator in the CI breaks when the image
+          # updates (since Visual Studio is also updated at the same time).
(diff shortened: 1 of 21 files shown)


imciner2:
The failures in the Windows CI tests are due to GitHub updating the image being used as `windows-latest` from a Windows 2019-based image with Visual Studio 2019 to a Windows 2022-based image with Visual Studio 2022. This broke our CMake generator logic, so I have updated our CI to hardcode the image and VS version to fix the Windows CI builds.

vineetbansal:
Looks good. Thanks @imciner2. Good to know about the VS upgrade on gh-actions too. That's probably the issue with my other branch that I haven't looked at for a few days.

PR #406, opened 2022-03-16
some interface fixes for update_matrices (for mkl algebra)

vineetbansal:
Since the signature of `update_linsys_solver_matrices` has changed, compilation with `-DALGEBRA=mkl` was failing. I looked at the recent changes in the `qdldl` case and am mirroring the changes here for the mkl interface. @goulart-paul - perhaps you can take a look and see if I missed something?

Also, for `-DEMBEDDED=2`, compilation was failing due to `NULL` being undefined. I've replaced those instances with `OSQP_NULL`. @imciner2 - is `0` is a better replacement here?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,6 +106,7 @@ jobs:
 
         - name: Valgrid check
           run: |
+            sudo apt-get update
             sudo apt-get install valgrind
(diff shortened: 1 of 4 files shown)


vineetbansal:
Looks like the only test failing is the valgrind check on Linux due to `apt install valgrind` currently not working - a problem with Azure cloud. I'll trigger this again later to see if that error goes away.

PR #419, opened 2022-04-13
Update auxil.c

vineetbansal:
Changed comment on dual infeasibility check from:
```
q * delta_x < eps * || delta_x ||
```
to
```
 q * delta_x < 0
```
The code is already doing this.


bstellato (approved):
Great! Thanks

PR #420, opened 2022-04-26
Refactor CI jobs and test more combinations

imciner2:
Instead of always installing MKL on the CI job, only install it in the ones that use the MKL algebra. This allows for the non-MKL jobs to continue even if there are issues installing it from conda.

### DIFF
--- a/.github/workflows/mkl.yaml
+++ b/.github/workflows/mkl.yaml
@@ -98,7 +98,6 @@ jobs:
                   -S . -B $OSQP_BUILD_DIR_PREFIX \
                   -DALGEBRA='mkl' \
                   -DFLOAT=OFF \
-                  -DEMBEDDED=0 \
                   -DPROFILING=ON \
                   -DCTRLC=ON \
                   -DPRINTING=ON \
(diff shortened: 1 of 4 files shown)


imciner2:
Whoops, I did not want to double the amount of CI jobs we run :grimacing:. Let me look into that.

.github/workflows/mkl.yaml
```diff
@@ -0,0 +1,116 @@
...
+        fail-fast: false
+
+        matrix:
+          # Specify the exact Windows image because the CMake generator in the CI breaks when the image
+          # updates (since Visual Studio is also updated at the same time).
+          os: [ubuntu-latest, macos-latest, windows-2022]
+          python-version: [3.9]
+          long: ['ON', 'OFF']
+
+          # MKL only supports double precision in the iterative solver, so that is the only combination we support as well.
```
vineetbansal, line 26: Perhaps we add `float: ['OFF']` as a single matrix entry to go right next to this comment and use it down below? Just so it remains close to the part of the CI if affects (currently).

vineetbansal:
Looks great - just the one comment, which is not all that important anyway. Thanks Ian!

.github/workflows/mkl.yaml
imciner2: Good idea, that way it better mimics the other CI workflows. Just pushed that change.

PR #430, opened 2022-06-21
Fixes to codegen to handle unconstrained problems

imciner2:
There were no codegen tests handling unconstrained problems before, so I have added some. This exposed some errors in how the data was written out (that caused problems both at writing time and at compilation time), so this now fixes these errors. It also moves the nonconvex tests into the codegen test header to consolidate them with the new unconstrained tests.


imciner2:
And this also exposed another issue with how empty CSC matrices were handled at the input. The setup routine was always allocating vectors for the row indices and values when it copied the input CSC matrix, even when the input had neither of those. This caused a memory access issue in the codegen (which was the valgrind failure). The most recent commit is an attempt to fix it by allowing the copy routine to not allocate those two vectors.

imciner2:
I have updated this PR to now also contain new tests for linear programs (i.e. no P matrix in the cost), and also for code generation of the linear programs to ensure they are written and compiled correctly.

imciner2:
It appears that the Cuda backend is failing on the new LP test. The solver reports an optimal solution, but the comparison between the expected solution. I haven't had a chance to dig into it to see if it is a problem with the iterative linear solver or something more fundamental.

bstellato (approved):
LGTM @imciner2. Thanks for these fixes.

imciner2:
Excellent. @bstellato thanks for looking at it, and thanks for identifying the cause of the test failure.

PR #433, opened 2022-07-05
Support appending the algebra type to the library name

imciner2:
By appending the algebra type to the soname, the users can then ship multiple shared libraries next to eachother, each with a different algebra backend.

If the user specified the CMake flag `-DOSQP_LIB_SUFFIX_ALGEBRA=ON`, it will create libraries that look like the following:

For the default algebra:
```
$ objdump -x libosqp-builtin.so | grep SONAME
  SONAME               libosqp-builtin.so
$ ls build/out
/home/b63738im/dev/optimization/osqp/c_code/build/out
Permissions Size User     Date Modified Git Name
.rwxrwxr-x@ 139k b63738im  5 Jul 14:59   -I libosqp-builtin.so
.rw-rw-r--@ 231k b63738im  5 Jul 14:59   -I libosqpstatic-builtin.a
.rwxrwxr-x@ 127k b63738im  5 Jul 14:59   -I osqp_codegen_demo
.rwxrwxr-x@ 127k b63738im  5 Jul 14:59   -I osqp_demo
```

For MKL:
```
$ objdump -x libosqp-mkl.so | grep SONAME
  SONAME               libosqp-mkl.so
$ ls build/out
/home/b63738im/dev/optimization/osqp/c_code/build-mkl/out
Permissions Size User     Date Modified Git Name
.rwxrwxr-x@  96k b63738im  5 Jul 14:51   -I libosqp-mkl.so
.rw-rw-r--@ 142k b63738im  5 Jul 14:51   -I libosqpstatic-mkl.a
.rwxrwxr-x@  92k b63738im  5 Jul 14:51   -I osqp_demo

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,11 @@ option(PRINTING "Enable solver printing" ON)
 option(PROFILING "Enable solver profiling (timing)" ON)
 option(CTRLC "Enable user interrupt (Ctrl-C)" ON)
 
+# Allow appending the algebra shortname to the end of the library and the soname so people can have
+# multiple libraries side-by-side on an install.
+option(OSQP_LIB_SUFFIX_ALGEBRA "Add the algebra name to the library name" OFF)
+mark_as_advanced(OSQP_LIB_SUFFIX_ALGEBRA)
+
 # Set the relevant boolean values for the algebra
 if(${ALGEBRA} STREQUAL "default")
   set(ALGEBRA_DEFAULT ON)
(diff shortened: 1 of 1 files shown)


imciner2:
> Looks like the flag is called OSQP_LIB_SUFFIX_ALGEBRA and not OSQP_LIB_SUFFIX_ALGEBRA_SUFFIX as in the thread. Just FYI in case you meant to use the latter.

Yea, I meant what was in the code but I guess I mistyped the comment.

> I thought the soname burned in the .so should have the major version number, and the filename itself a major.minor format? I may have outdated information, and I've never seen the version information with osqp libraries anyway. Again, just mentioning this in case you think it's worth opening up an issue for the future..

If we want we can add a soversion suffix to the files, which will be the API version rather than the actual version of the library. It is fairly easy for me to add that in a new PR, so I will do that later and see about merging this first (it is needed for some of my work on the Julia interface).

Any further comments, or does this look good to merge?

vineetbansal:
Thanks @imciner2 !

PR #437, opened 2022-07-23
Add function to query capabilities of the current library and abstract linear solver information

imciner2:
Adding a function to query the capabilities of the solver allows the user code/interfaces to query the solver for what it supports instead of having to hard-code specific information. This also allows the unit test suite to use this information to choose which linear system solvers to test automatically (with no need for define statements in the testing code).

I also abstracted the default linsys into the algebra, so now a new custom algebra a user develops can choose what solver to use without having to modify the actual solver functions.

@vineetbansal Once this is merged and you merge the develop branch into your derivative branch, you should add a new capability for derivative computation (since that will only be available on the builtin algebra at first).

PR #438, opened 2022-07-25
Clean up algebra linking and always link against MKL

imciner2:
Even on Linux we need to tell the shared library it has a dependency on MKL, otherwise the dynamic loader can have issues with undefined symbols because it doesn't know to load the MKL library. I have also cleaned up the linking so now the algebras define the libraries in a similar way to the include directories, removing the need for all the conditionals in the CMake code (and making the algebras more self contained).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,11 @@ get_property(
   TARGET OSQPLIB
   PROPERTY SOURCES)
 
+get_property(
+  osqplib_link_libs
+  TARGET OSQPLIB
+  PROPERTY LINK_LIBRARIES)
+
 # ----------------------------------------------
 # Language-specific compilation checks
 # ----------------------------------------------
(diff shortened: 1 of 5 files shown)


imciner2:
This has ballooned a bit and now also changes the linkage for MKL from being to the individual dynamic library to the single dynamic library and then initialize the integer interface before first use of the algebra. This is needed to get the MKL backend to work in Julia's binary ecosystem because it is unable to load the individual library properly, but can load the single dynamic library (this is also the way the Julia MKL.jl interface to MKL operates).

I also had to fix the signature for the algebra init function for MKL because it was missing the argument. This meant that it was never getting called properly, and the new setting of the integer type was broken without this fix.

imciner2:
This and #437 have some git conflicts when merged together, but they can be reviewed independently since they are functionally different (it is just git doesn't like merging the algebra file).

vineetbansal:
LGTM. Do you know how setting the mkl interface in algebra/mkl/CMakeLists.txt:
```
if(DLONG)
  set(MKL_INTERFACE "ilp64")
else()
  set(MKL_INTERFACE "lp64")
endif()
```
differs from what you're doing now in osqp_algebra_init_libs:
```
#ifdef DLONG
    retval = mkl_set_interface_layer(MKL_INTERFACE_ILP64);
#else
    retval = mkl_set_interface_layer(MKL_INTERFACE_LP64);
#endif
```
Are both needed?

imciner2:
> differs from what you're doing now in osqp_algebra_init_libs:

The difference is in the library file that it links against after compilation. When setting it inside the `osqp_algebra_init_libs`, it should link against the single dynamic library (SDL) `libmkl_rt.so` instead of the integer-specific library. This function call then tells that library which functions to call based on the integer requested. When cmake is set to not use the SDL mode, then the `MKL_INTERFACE` variable tells cmake which library file to choose on the link line, making the executable link against the very specific one for that integer type.

In the case of Julia, they only load the generic library and then call into the library to set the integer type before use. So when we linked to the specific library, it couldn't be found in the Julia package path and the loading failed.

> Are both needed?

Technically, we don't need the CMake portion setting `MKL_INTERFACE` anymore, since it doesn't actually do anything in `SDL` mode, but I left it in there so I could more easily remember which interface was needed when.

PR #440, opened 2022-07-31
Implement BLAS level 2 functions in the MKL algebra backend

imciner2:
This adds the calls to the BLAS level 2 MKL sparse operations for the MKL backend. It turns out that the only operations it really implements that we use are the matrix-vector multiplication. The actual memory that backs up the opaque MKL matrix object can be stored in our existing csc matrix object, so we use that as a shadow object to store the data. We can also perform many of the operations that would be very difficult using MKL on the shadow csc object and have them transparently appear in the MKL matrix (such as the matrix scaling, left/right diagonal multiplication and the subrow extraction). While technically not documented as being a supported method (they leave the type opaque and don't technically guarantee they won't change it, I don't think they are likely to change the memory management strategy since that would mean probably copying the user data into a new array (which would be very expensive for large matrices, so not a useful change).

There is still an issue with this though, and that is the linear algebra unit test is failing on the matrix-vector multiplication test with the new MKL function:
```
./out/osqp_tester test_lin_alg
Filters: test_lin_alg

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
osqp_tester is a Catch v2.13.6 host application.
Run with -? for options

-------------------------------------------------------------------------------
test_lin_alg
  test_mat_vec_multiplication
-------------------------------------------------------------------------------
/home/imcinerney/dev/osqp/code/master/tests/osqp_tester.cpp:39
...............................................................................

/home/imcinerney/dev/osqp/code/master/tests/lin_alg/test_lin_alg.h:199: FAILED:
  REQUIRE( OSQPVectorf_norm_inf_diff(result.get(), ref.get()) < 1e-4 )
with expansion:
  0.1507309222 < 0.0001
with message:
  Linear algebra tests: error in matrix-vector operation, matrix-vector
  multiplication

===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 12 | 11 passed | 1 failed
```

I haven't had a chance to fully debug this error though.


imciner2:
Ok, I think this is ready now. I figured out why the matrix-vector multiplication was failing before, and it was because I was overcomplicating the MKL CSC format and ours, so there was an off-by-one error in the column pointer vector. With that fixed the matrix-vector multiplication works now.

I ran into the most trouble with the empty matrices actually, because MKL was very specific about how the vectors should be passed in. So I created a new suite of unit tests just for empty/all zero matrices and vectors to ensure it all works. That test suite passes on all three algebras.

@bstellato I looked into using more MKL operations for more of the algebra operations we have, and I don't think it is actually feasible. The problem with using the matrix-matrix routines for the diagonal multiplication and the scaling operation is that there are only a few different types, but they all do stuff we don't want:
* `mkl_sparse_d_mm` to compute `C = alpha*op(A)*B + beta*C` - While tempting to use this for scaling by just doing `beta*C`, it turns out that `C` is a dense matrix, so we can't use our sparse matrix there (and it then also stores the result in a dense matrix, so we lose sparsity).
* `mkl_sparse_spmm` to compute `C = op(A) *B` - While this does use `A` and `B` as sparse matrices, it will allocate a brand new matrix `C` for the result, which we don't want (we will have the same non-zero pattern and want to just overwrite the existing matrix, so this is very inefficient).
* `mkl_sparse_d_spmmd` to compute `C = op(A)*B` - This takes `A` and `B` as sparse matrices, but the result `C` is a dense matrix - bad for us.
* `mkl_sparse_sp2m` to compute `C = opA(A) *opB(B)`- This takes two sparse matrices and then allocates the result as a new sparse matrix (the difference is this one allows transposing either/both operands).
* `mkl_sparse_d_sp2md` to compute `C = alpha * opA(A) *opB(B) + beta*C` - Similar to `mkl_sparse_d_mm` with a dense result, except now you can transpose both operands.
So basically, none of the matrix-matrix multiplication routines will actually be efficient for us, so I think the current manipulation of the backing CSC matrices is the best choice for these operations. This does mean that MKL is really only used in the matrix-vector multiplies, so we will probably see the most speed up from these changes in the indirect (PCG) solver.

@vineetbansal do you think you could give this PR a test on the benchmarks to see if you can compare it to the previous non-MKL BLAS level 2 set that you did?

bstellato:
That's awesome, @imciner2! Thank you for fixing the indexing. Since there were no segfaults, I suspected it was related to the indexing in MKL. I agree other O(n)-type MKL operations are not feasible and that it would not bring noticeable speedups. After @vineetbansal's benchmarks, I'd be happy to merge this PR.

PR #445, opened 2022-08-06
Enable user-defined functions for interrupt and timing interfacing

imciner2:
This changeset cleans up the existing custom memory and printing code to remove the additional two headers that were added, and instead include it from existing headers. This allows for their use in embedded mode easier.

The interrupt system has been adapted to allow for people to add their own implementations for the three functions inside `interrupt.h` now, and this is used to add the ctrl-c handlers for the various platforms as well.

The timing system has been adapted to support user implementations of timers, and has also changed to an opaque type for the `OSQPTimer` type in the main solver. This allows for the timer implementations to have an entire creation routine to setup the timer as needed (e.g. allocate the memory for the timer structure). The actual codegen system needs to be extended to support writing out workspaces with profiling enabled (e.g. make the structs have the proper fields added for that).

I have also gone through and started removing all the interface-specific code in the library, since these parts are the last bits that contain interface-specific code. I have updated the issues in the interface repos (https://github.com/osqp/osqp-python/issues/76, https://github.com/osqp/osqp-matlab/issues/47, https://github.com/osqp/osqp-r/issues/28) with the code that was removed from the main library, and some details on how to add it to the interface build. A future PR will actually strip out the interface search methods for MATLAB/R/Python from the CMake build system, finishing the work to separate the interfaces from the main solver library.


vineetbansal:
@imciner2 - this looks fine to me, since I'm concerning myself with https://github.com/osqp/osqp-python/issues/76 and the tests all pass here. I feel like I'm not qualified to comment on the actual changes here, but let me know if you're satisfied and I can merge this (and continue on the Python wrapper side of things).

imciner2:
@vineetbansal thanks for the quick look. I have merged this now, so you can start modifying the Python interface to work with the new systems. Let me know if you have any problems/issues with doing that.

PR #449, opened 2022-08-16
Cleanups to the algebra interfaces/structure

imciner2:
This PR contains various cleanups to the algebra interface/build structure that makes it more friendly to users and custom algebras. Specifically, it contains the following:
* Moved QDLDL into a new `_common/lin_sys/qdldl` directory since that solver can be used by any linear algebra backend, since it doesn't rely on a specific library being loaded.
* Renamed the CMake options and the defines for the algebras to be namespaced. The main change is that now CMake wants `-DOSQP_ALGEBRA_BACKEND=` instead of `-DALGEBRA`. This also namespaces the algebra defines in our configure header so they won't conflict with ones in user code.
* Renamed the default algebra to be `builtin` everywhere
* Made the algebras responsible for their name instead of the main CMake file. That function should be trivial in all implementations, so the compiler is likely to do constant folding/optimizations to it.


vineetbansal (approved):
Thanks @imciner2 - changes make sense. Since some critical cmake variable names/values have been renamed I guess we should merge this without much delay, unless @bstellato has any concerns.

imciner2:
Seeing no further comments, I have just merged this so we don't encounter any issues with the refactoring becoming stale/causing merge conflicts.

PR #453, opened 2022-08-25
Fix linkage of generated solver object

imciner2:
The generated solver object should have C linkage so that C++ compilers don't mangle its name and then prevent linking against it. This doesn't appear to be a problem for some solvers (like GCC) because they only mangle the names that could possibly have multiple definitions (like static variables and functions), whereas MSVC appears to mangle all names when compiled in a C++ context.

@vineetbansal this should fix the compilation error in the generated Python wrapper you showed me earlier.

(the main change is in the second commit, the first just adds a new test executable compiled with a C++ compiler to make sure the include works in both compilers).

### DIFF
--- a/.github/workflows/codegen.yaml
+++ b/.github/workflows/codegen.yaml
@@ -45,13 +45,23 @@ jobs:
         # Test that embedded mode 1 works
         - name: Embedded mode 1 (vector)
           run: |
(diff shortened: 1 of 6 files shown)


vineetbansal (approved):
Looks good. Thanks @imciner2 !

PR #454, opened 2022-08-25
Add dedicated flag in CMake and codegen for derivative support

imciner2:
This makes derivative support behave logically the same as the codegen support, where it can be enabled/disabled if the user chooses. This will also allow for easier enablement of the derivatives on platforms in the future (such as through adding a non-QDLDL solver like a custom LSQR to enable it on other algebra backends, or making it codegen friendly).

This also adds a flag to the codegen defines struct that is supposed to enable/disable the derivatives, but this actually isn't going to do anything yet though since we can't do derivatives in embedded until we create a static workspace for it.


vineetbansal (commented):
Thanks @imciner2 !

PR #456, opened 2022-08-25
Rename base number/matrix types

imciner2:
This is the rename of the base `c_float` and `c_int` types along with the `csc` type to become `OSQPFloat`, `OSQPInt` and `OSQPCscMatrix`, respectively.


imciner2:
@vineetbansal this does a rename for the CMake to choose long/floats and the embedded mode, so once it is merged the Python interface will have to be modified accordingly.

PR #469, opened 2022-09-10
Pre-commit

vineetbansal:
A simple `pre-commit` starter configuration that currently checks for `yml` errors. We have a few `yml` files in this repo, and all currently pass the checks.

Use `pre-commit install` when in the `osqp` checkout folder to enable automated checks. For the CI, installing `pre-commit` through `conda` from the `conda-forge` channel seems to me to be a reasonable option. For local development when you don't want to use `conda` unnecessarily, I see that Ubuntu has a `apt install pre-commit` available. Other distros might have it too.

I'll add some python-related hooks in a future PR. C-related hooks seem to be at:
https://github.com/pocc/pre-commit-hooks

No modifications to the new gh-workflow file should be needed as we add more hooks to `.pre-commit-config.yaml` in the root folder.

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -1,0 +1,39 @@
+name: Pre-commit checks
+
+on:
+  push:
+    branches: [ master, develop**, vb/precommit ]
+    tags:
+      - '*'
+  pull_request:
+    branches: [ master, develop** ]
+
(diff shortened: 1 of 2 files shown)


imciner2:
Thanks. We should definitely document how to do this in the contributing file we need to write.

PR #470, opened 2022-09-10
Renamed osqp_capabilities enum to osqp_capabilities_type

vineetbansal:
I'm starting to incorporate supported capabilities in the python wrapper so as to better handle unsupported calls at runtime. I noticed that this renaming needs to happen for to avoid naming conflicts when compiling in the wrapper (since there's an `osqp_capabilities` api function already).
 
 There seems to be no place in the `osqp` codebase itself where this renaming affects anything though.

### DIFF
--- a/include/public/osqp_api_constants.h
+++ b/include/public/osqp_api_constants.h
@@ -6,7 +6,7 @@
 /***********************
 * Solver capabilities *
 ***********************/
-enum osqp_capabilities {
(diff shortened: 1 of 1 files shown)


imciner2 (approved):

imciner2:
Thanks for fixing this. I hadn't thought about possible name collisions when I wrote this.

PR #471, opened 2022-09-10
Fix spelling errors in capability defines

imciner2:
I made a spelling mistake in the original implementation of the capability definitions, so this fixes it in all places now.


vineetbansal:
Ah thanks - I was copy-pasting these constants all morning and never noticed the typo!

PR #472, opened 2022-09-12
Building docs for develop-1.0 in a subfolder

vineetbansal:
This pair of steps (which **have** to run in the same workflow action as the main documentation), check out the `develop-1.0` branch of the repo, and build its documentation in a `develop-1.0` folder inside the current `docs` folder on the target `gh-pages` branch. I've tested this workflow in another repo and it seems to work well (and is the only way to do this AFAIK).

If all goes well, we should see, in addition to:
https://osqp.org/docs/

a parallel link:
https://osqp.org/docs/develop-1.0

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,7 @@ name: Docs
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, develop-1.0 ]
     tags:
       - '*'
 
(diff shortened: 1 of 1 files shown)


imciner2 (approved):
This seems good, so lets deploy it and see how the docs look.

imciner2:
Although, I see this is going towards the master branch, so how will we regenerate the docs as we develop the new develop-1.0 docs? Can we trigger this workflow separately, or do we need to merge master into develop-1.0 and have this triggered each time?

vineetbansal:
The workflow is triggered on push to either `master` or `develop-1.0`. In either case, the main documentation is built by checking out the `master` branch, and additional documentation (in a `develop-1.0` subfolder of `docs`) is built by checking out the `develop-1.0` branch in addition to the main `master` checkout.

There was an oversight in the workflow, unrelated to your concern here, that I've fixed in a subsequent commit.

vineetbansal:
I will also send a parallel PR towards the `master` branch for `docs.yml`, since in its absence, any pushes to `master` while `develop-1.0` is still unmerged will wipe out the `develop-1.0` documentation in the target branch (`gh-pages`). That is the branch where the built/rendered html lives.

imciner2:
Yep, that other PR fixes my concerns and will allow us to actually develop the new docs. I'll merge this one as well.

PR #473, opened 2022-09-15
Skip polishing when there is no active set detected

imciner2:
Polishing really only makes logical sense when there is an active set to polish against, so to prevent issues with the linear algebra when the A matrix is 0, just don't polish the solution if there is no active set. This is shown to the user through a notification and a new return status for the solution polishing variable:

```
-----------------------------------------------------------------
        OSQP v0.0.0  -  Operator Splitting QP Solver
           (c) Bartolomeo Stellato,  Goran Banjac
     University of Oxford  -  Stanford University 2021
-----------------------------------------------------------------
problem:  variables n = 5, constraints m = 0
       nnz(P) + nnz(A) = 5
settings: algebra = Built-in,
       linear system solver = QDLDL v0.1.6,
       eps_abs = 1.0e-05, eps_rel = 1.0e-05,
       eps_prim_inf = 1.0e-04, eps_dual_inf = 1.0e-04,
       rho = 1.00e-01 (adaptive),
       sigma = 1.00e-06, alpha = 1.60, max_iter = 4000
       check_termination: on (interval 25),
       time_limit: 1.00e+10 sec,
       scaling: on, scaled_termination: off
       warm starting: on, polishing: on, 
iter   objective    prim res   dual res   rho        time
1  -1.2294e+01   0.00e+00   1.42e+00   1.00e-01   8.11e-05s
25  -1.9210e+01   0.00e+00   6.72e-06   1.00e-01   1.05e-04s
Polishing skipped - no active set detected at optimal point

status:               solved
solution polishing:   skipped
number of iterations: 25
optimal objective:    -19.2098
run time:             1.11e-04s
optimal rho estimate: 1.00e-06
```

This PR also renames the `OSQPVectorf_mean` function to `OSQPVectorf_pos_mean` to make clear there is an assumption that all values of the vector are positive. This is because assuming all positive values allows us to optimize both BLAS_based backends and the builtin backend (BLAS backends can call `asum` and the builtin can avoid a call to `abs`). We can do this because we only ever want the average of a positive vector - since the vector contains the infinity norms of the matrix columns of P.

### DIFF
--- a/include/private/algebra_vector.h
+++ b/include/private/algebra_vector.h
@@ -166,8 +166,8 @@ OSQPFloat OSQPVectorf_norm_inf_diff(const OSQPVectorf* a,
 /* ||v||2 */
 OSQPFloat OSQPVectorf_norm_2(const OSQPVectorf* v);
 
-/* mean of vector elements - assumes all elements are positive */
-OSQPFloat OSQPVectorf_pos_mean(const OSQPVectorf* a);
+/* ||v||1 */
+OSQPFloat OSQPVectorf_norm_1(const OSQPVectorf* a);
 
 /* Inner product a'b */
 OSQPFloat OSQPVectorf_dot_prod(const OSQPVectorf* a,
@@ -224,9 +224,6 @@ OSQPInt OSQPVectorf_in_reccone(const OSQPVectorf* y,
                                OSQPFloat          tol);
 
 # if OSQP_EMBEDDED_MODE != 1
-
-/* Vector mean value*/
-OSQPFloat OSQPVectorf_pos_mean(const OSQPVectorf* a);
 
 /* Vector elementwise reciprocal b = 1./a (needed for scaling)*/
 void OSQPVectorf_ew_reciprocal(OSQPVectorf*       b,
--- a/include/private/algebra_vector.h
+++ b/include/private/algebra_vector.h
@@ -166,8 +166,8 @@ OSQPFloat OSQPVectorf_norm_inf_diff(const OSQPVectorf* a,
 /* ||v||2 */
 OSQPFloat OSQPVectorf_norm_2(const OSQPVectorf* v);
 
-/* mean of vector elements */
-OSQPFloat OSQPVectorf_mean(const OSQPVectorf* a);
+/* mean of vector elements - assumes all elements are positive */
+OSQPFloat OSQPVectorf_pos_mean(const OSQPVectorf* a);
 
 /* Inner product a'b */
 OSQPFloat OSQPVectorf_dot_prod(const OSQPVectorf* a,
@@ -226,7 +226,7 @@ OSQPInt OSQPVectorf_in_reccone(const OSQPVectorf* y,
 # if OSQP_EMBEDDED_MODE != 1
 
 /* Vector mean value*/
-OSQPFloat OSQPVectorf_mean(const OSQPVectorf* a);
+OSQPFloat OSQPVectorf_pos_mean(const OSQPVectorf* a);
 
 /* Vector elementwise reciprocal b = 1./a (needed for scaling)*/
 void OSQPVectorf_ew_reciprocal(OSQPVectorf*       b,
--- a/src/osqp_api.c
+++ b/src/osqp_api.c
@@ -320,7 +320,7 @@ OSQPInt osqp_setup(OSQPSolver**         solverp,
     return osqp_error(OSQP_MEM_ALLOC_ERROR);
 
   // Initialize information
-  solver->info->status_polish = 0;              // Polishing not performed
+  solver->info->status_polish = OSQP_POLISHING_UNPERFORMED; // Polishing not performed
   update_status(solver->info, OSQP_UNSOLVED);
 # ifdef OSQP_ENABLE_PROFILING
   solver->info->solve_time  = 0.0;                   // Solve time to zero
--- a/src/osqp_api.c
+++ b/src/osqp_api.c
@@ -320,7 +320,7 @@ OSQPInt osqp_setup(OSQPSolver**         solverp,
     return osqp_error(OSQP_MEM_ALLOC_ERROR);
 
   // Initialize information
-  solver->info->status_polish = OSQP_POLISHING_UNPERFORMED; // Polishing not performed
+  solver->info->status_polish = OSQP_POLISH_NOT_PERFORMED; // Polishing not performed
   update_status(solver->info, OSQP_UNSOLVED);
 # ifdef OSQP_ENABLE_PROFILING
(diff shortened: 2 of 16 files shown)


src/util.c
```diff
@@ -212,10 +212,12 @@ void print_footer(OSQPInfo* info,
   c_print("status:               %s\n", info->status);
 
   if (polishing && (info->status_val == OSQP_SOLVED)) {
-    if (info->status_polish == 1) {
+    if (info->status_polish == OSQP_POLISH_SUCCESS) {
       c_print("solution polishing:   successful\n");
     } else if (info->status_polish < 0) {
       c_print("solution polishing:   unsuccessful\n");
+    } else if (info->status_polish == OSQP_POLISH_NO_ACTIVE_SET) {
+      c_print("solution polishing:   skipped\n");
```
goulart-paul, line 220: I think calling this "skipped" could cause confusion.   If polishing is disabled as an option then it is also "skipped", and user might reasonably ask why the polishing step was skipped when they had it enabled.

Maybe "not needed" would be clearer here?

src/polish.c
```diff
@@ -296,8 +297,17 @@ OSQPInt polish(OSQPSolver* solver) {
   mred = form_Ared(work);
   if (mred < 0) {
     // Polishing failed
-    info->status_polish = -1;
-    return -1;
+    info->status_polish = OSQP_POLISH_FAILED;
+    return OSQP_POLISH_FAILED;
+  } else if (mred == 0) {
+    /* No active constraints, so skip polishing */
+    c_print("Polishing skipped - no active set detected at optimal point\n");
```
goulart-paul, line 304: "skipped" -> "not needed" is clearer I think.

include/private/algebra_vector.h
```diff
@@ -166,8 +166,8 @@ OSQPFloat OSQPVectorf_norm_inf_diff(const OSQPVectorf* a,
 /* ||v||2 */
 OSQPFloat OSQPVectorf_norm_2(const OSQPVectorf* v);
 
-/* mean of vector elements */
-OSQPFloat OSQPVectorf_mean(const OSQPVectorf* a);
+/* mean of vector elements - assumes all elements are positive */
```
goulart-paul, line 169: I don't understand what the assumptions are behind the `pos_mean` method here.   Is it "mean assuming all entries" are positive?  Is it "mean of the absolute values"?    If the latter, why not just take the one norm over the length?

imciner2: The assumption is that all entries in `a` are positive, which will allow the mean to be implemented one of two ways and get equivalent results:
1) Using the one norm divided by the length
2) Using the normal mean formula

Method 1 is done in the MKL and CUDA backend by using the `asum` class of BLAS functions followed by the divide by the length, and method 2 is used in the builtin backend because an equivalent 1-norm computation would involve the absolute value of the elements of the vector (so requires another operation/temp variable on each element). This actually was a discrepancy between the CUDA and MKL/builtin algebras before, with CUDA implicitly doing option 1 and MKL/builtin doing option 2, but all three were advertising they were doing the mean. So the idea behind this is to codify that assumption into the API, which also allows MKL to now call the BLAS function instead of doing a loop through all elements of the vector.

This assumption is valid in the place we use the mean - the matrix scaling computation - since it is taking the mean over the result of the infinity norms of the columns and the infinity norm is already defined to be always positive.

If we wanted to implement the true mean computation, it would involve writing a custom kernel to compute the sum of the vector entries in CUDA, and means we won't take advantage of the pre-made BLAS functions and their potential speed-up in either CUDA or MKL. Doing a true "absolute mean" (which is the mean over the absolute values of the vector elements) would similarly seem to slow down the builtin algebra computation due to the added absolute value operation.

src/util.c
imciner2: Should we then change the reported status to be `OSQP_POLISH_NOT_NEEDED` instead of saying `OSQP_POLISH_NO_ACTIVE_SET` to make it clear in the error code as well? (the current one is advantageous because it says why we didn't polish, but I guess on reading it it isn't actually clear we didn't polish).

include/private/algebra_vector.h
goulart-paul: Is the issue here down to BLAS not providing a direct way to sum a vector?  If so, then I think it is possible to do it using the BLAS `dot` function to compute a sum of elements in a vector `x` by telling it to compute <x , y> and then specifying that `y = 1.0` with an increment of `0`.

In that way you could directly compute a bare sum in MKL or wherever to compute the actual mean, and then just use that everywhere.   It does not seem like good design to have the actual behaviour of the function change depending on the algebra backend, even if it is only ever supplied with nonnegative inputs.

imciner2: > Is the issue here down to BLAS not providing a direct way to sum a vector?

Basically, yes - BLAS provides no way to directly sum a single vector. My concern with the dot function is that we are then adding all those multiplication operations (one per addition), which I guess might be only 1-2 cycles on most machines, but that could add up for larger problems, which is what the MKL and CUDA backends target. If we truly don't want to rely on the API specifying the vector is all positive, then adding an abs to the builtin algebra is probably the better option.

goulart-paul: Isn't this mean function only used in the equilibration step anyway?   That's already an extremely small part of the total compute time, so I don't think any extra ops here matter.

If you're going to add an abs to the building algebra anyway, then at that point we might as well drop this function entirely and manually compute in the 1-norm / length in the one place (I think) it is used.

src/util.c
imciner2: Cleaned up in most recent commit.

src/polish.c
imciner2: Change to "not needed" in most recent commit.

src/osqp_api.c
```diff
@@ -320,7 +320,7 @@ OSQPInt osqp_setup(OSQPSolver**         solverp,
     return osqp_error(OSQP_MEM_ALLOC_ERROR);
 
   // Initialize information
-  solver->info->status_polish = 0;              // Polishing not performed
+  solver->info->status_polish = OSQP_POLISHING_UNPERFORMED; // Polishing not performed
```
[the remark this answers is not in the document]
imciner2: Change to `OSQP_POLISH_NOT_PERFORMED` to be clearer and also match the naming of the other status values for polishing.

include/private/algebra_vector.h
imciner2: I have gone ahead and made the algebra libraries compute the 1-norm using the asum functions (if available) and then the usage of the average actually computes the mean at the location it is needed.

goulart-paul (approved):

PR #475, opened 2022-09-20
Allow CUDA to be fully statically linked

imciner2:
This adds a new option `OSQP_CUDA_STATIC_LINKAGE` that will statically link against the CUDA numerical libraries, meaning all CUDA libraries are now statically linked. This removes the need for a CUDA runtime on the computer that runs the library.


imciner2:
Renamed the option to be `OSQP_CUDA_STATIC_LINKING` and added a warning that this option will create extremely large libraries (my experiments show the library goes from 1.3MB to 330MB with static linking of CUDA). We'll merge this once the majority of CI passes.

imciner2:
CUDA error is unrelated to this PR and occurs in other places, so merging this in.

PR #476, opened 2022-09-23
CI: Run all tests compiled for a configuration

imciner2:
Instead of individually calling the tester, just use the ctest wrapper to call all the tests. Before this change we were skipping the linear algebra tests because they are now a separate tester executable, so with this change we restore the linear algebra tests.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -33,6 +33,7 @@ jobs:
 
       env:
         OSQP_BUILD_DIR_PREFIX: ${{ github.workspace }}/build
+        CTEST_OUTPUT_ON_FAILURE: 1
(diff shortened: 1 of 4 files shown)


vineetbansal:
Looks great! Thanks @imciner2. Looks like in our case we need to do a plain `make` before a `make test` will succeed (as we're doing here in the CI). Just pointing it out in case you think this is something that needs "fixing".

imciner2:
> Looks like in our case we need to do a plain make before a make test will succeed (as we're doing here in the CI). Just pointing it out in case you think this is something that needs "fixing".

I think that's probably fine. We could try to update the CMake to automatically add the dependency there, but it is lower on the priority list.

PR #482, opened 2022-10-13
Vb/adjoint split

vineetbansal:
The `osqp_adjoint_derivative` api function is now gone in favor of the following 3:
  - `osqp_adjoint_derivative_compute`
  - `osqp_adjoint_derivative_get_mat`
  - `osqp_adjoint_derivative_get_vec`

I'm saving the essential stuff I need between the 1st and other calls in a new `OSQPDerivativeData` struct referenced by `OSQPWorkspace`.

This is more of an api introduction rather than making sure that we're not doing extra computation in `osqp_adjoint_derivative_compute` if `osqp_adjoint_derivative_get_mat` is never called, for example.

I based this branch off of `im/precond` since that has a bunch of changes. You'll want to do a squash and merge here since this has a lot of useless commits.


include/public/osqp_api_functions.h
```diff
@@ -259,15 +259,20 @@ OSQP_API OSQPInt osqp_update_rho(OSQPSolver* solver,
...
-OSQP_API OSQPInt osqp_adjoint_derivative(OSQPSolver*    solver,
-                                         OSQPFloat*     dx,
-                                         OSQPFloat*     dy_l,
-                                         OSQPFloat*     dy_u,
-                                         OSQPCscMatrix* dP,
-                                         OSQPFloat*     dq,
-                                         OSQPCscMatrix* dA,
-                                         OSQPFloat*     dl,
-                                         OSQPFloat*     du);
+OSQP_API OSQPInt osqp_adjoint_derivative_compute(OSQPSolver*    solver,
```
imciner2, line 262: It would be great if you could add the Doxygen-style comments to these three functions as well to make clear the inputs/outputs of the functions and what each argument does.

include/public/osqp_api_functions.h
```diff
@@ -259,15 +259,20 @@ OSQP_API OSQPInt osqp_update_rho(OSQPSolver* solver,
...
+                                                 OSQPFloat*     dx,
+                                                 OSQPFloat*     dy_l,
+                                                 OSQPFloat*     dy_u);
+
+OSQP_API OSQPInt osqp_adjoint_derivative_get_mat(OSQPSolver*    solver,
```
imciner2, line 267: Do we need these functions to return ints? The code looks to always be returning 0.

src/osqp_api.c
```diff
@@ -1273,28 +1288,42 @@ void csc_set_data(OSQPCscMatrix* M,
...
-                                OSQPCscMatrix* dA,
-                                OSQPFloat*     dl,
-                                OSQPFloat*     du) {
-
-    OSQPInt status = adjoint_derivative(
+OSQPInt osqp_adjoint_derivative_compute(OSQPSolver*    solver,
+                                        OSQPFloat*     dx,
+                                        OSQPFloat*     dy_l,
+                                        OSQPFloat*     dy_u) {
+
```
imciner2, line 1295: We should add this code to the beginning of the three derivative functions to make sure they don't crash if called before setup.

```
  // Check if solver has been initialized
  if (!solver || !solver->work || !solver->work->derivative_data)
    return osqp_error(OSQP_WORKSPACE_NOT_INIT_ERROR);
```

src/derivative.c
```diff
@@ -170,20 +185,24 @@ OSQPInt adjoint_derivative(OSQPSolver*    solver,
...
 
     // --------- lambda
     OSQPVectorf* m_zeros = OSQPVectorf_malloc(m);
     OSQPVectorf_set_scalar(m_zeros, 0);
-    OSQPVectorf* y_u = OSQPVectorf_malloc(m);
-    OSQPVectorf_ew_max_vec(y_u, y, m_zeros);
-    OSQPVectorf* y_l = OSQPVectorf_malloc(m);
-    OSQPVectorf_ew_min_vec(y_l, y, m_zeros);
-    OSQPVectorf_mult_scalar(y_l, -1);
+    derivative_data->y_u = OSQPVectorf_malloc(m);
```
imciner2, line 196: Since `m` is known at setup time, we should allocate these inside `osqp_setup` alongside the overall `derivative_data` struct. Anything we free in `osqp_cleanup` should be allocated by setup for better traceability (and to maintain consistent code style).

src/derivative.c
```diff
@@ -170,20 +185,24 @@ OSQPInt adjoint_derivative(OSQPSolver*    solver,
...
+    OSQPVectorf_ew_max_vec(derivative_data->y_u, y, m_zeros);
+    derivative_data->y_l = OSQPVectorf_malloc(m);
```
imciner2, line 198: Same here, allocate in `osqp_setup`.

src/derivative.c
```diff
@@ -244,117 +266,59 @@ OSQPInt adjoint_derivative(OSQPSolver*    solver,
...
 
-    OSQPVectorf *ryl = OSQPVectorf_new(r_yl, m);
+    derivative_data->ryl = OSQPVectorf_new(r_yl, m);
     c_free(r_yl);
-    OSQPVectorf_ew_prod(ryl, ryl, y_l);
-    OSQPVectorf_mult_scalar(ryl, -1);
-    OSQPVectorf *ryu = OSQPVectorf_new(r_yu, m);
+    OSQPVectorf_ew_prod(derivative_data->ryl, derivative_data->ryl, derivative_data->y_l);
+    OSQPVectorf_mult_scalar(derivative_data->ryl, -1);
+    derivative_data->ryu = OSQPVectorf_new(r_yu, m);
```
imciner2, line 321: Allocate `derivative_data->ryu` in `osqp_setup` and just overwrite the data in that vector with the new values here.

src/derivative.c
```diff
@@ -244,117 +266,59 @@ OSQPInt adjoint_derivative(OSQPSolver*    solver,
...
         } else {
             r_yl[eq_indices_vec[j]] = -rhs_data[pos+j] / y_data[eq_indices_vec[j]];
             r_yu[eq_indices_vec[j]] = 0;
         }
-
     }
+    c_free(nu_sign_vec);
 
-    OSQPVectorf *ryl = OSQPVectorf_new(r_yl, m);
+    derivative_data->ryl = OSQPVectorf_new(r_yl, m);
```
imciner2, line 317: Allocate in `osqp_setup` and overwrite the data here.

src/osqp_api.c
```diff
@@ -855,6 +861,15 @@ OSQPInt osqp_cleanup(OSQPSolver* solver) {
     if (work->timer) OSQPTimer_free(work->timer);
 # endif /* ifdef OSQP_ENABLE_PROFILING */
 
+# ifdef OSQP_ENABLE_DERIVATIVES
+      if (work->derivative_data){
+          if (work->derivative_data->y_l) OSQPVectorf_free(work->derivative_data->y_l);
+          if (work->derivative_data->y_u) OSQPVectorf_free(work->derivative_data->y_u);
+          if (work->derivative_data->rhs) OSQPVectorf_free(work->derivative_data->rhs);
+          c_free(work->derivative_data);
```
imciner2, line 869: It doesn't look like this is freeing all the vectors that are contained in the `derivative_data` struct. Please make sure to free all data.

src/derivative.c
```diff
@@ -244,117 +266,59 @@ OSQPInt adjoint_derivative(OSQPSolver*    solver,
     OSQPVectorf_free(rhs_temp2);
     OSQPVectorf_free(zeros);
 
-    // ----------- Check
-    adjoint_derivative_linsys_solver(solver, settings, P_full, G, A_eq, GDiagLambda, slacks, rhs);
+    // --------- stuff things into derivative_data
+    // All this should go inside _compute
+    derivative_data->rhs = OSQPVectorf_copy_new(rhs);
```
imciner2, line 271: Allocate in `osqp_setup` and then just overwrite with the new data here.

imciner2 (changes requested):
Thanks for splitting the API. Overall the API looks good, but it just needs the comments in the header file saying what each argument is.

In terms of the code split, there is a bit of confusion about where the solver-level memory is being allocated/free'd, with it being allocated in the individual functions, but free'd globally. It is much better to allocate in the `osqp_setup` function (since we know the size of each vector in setup), and then free in `osqp_cleanup`.

include/public/osqp_api_functions.h
vineetbansal: I think in this case, since the implementation for these derivative functions will likely get more robust with time, it makes sense so that in the future we can return a suitable OSQP_* error under certain conditions, without changing the signature. What do you think?

src/osqp_api.c
vineetbansal: Yes - it looks like I don't free up `ryl` and `ryu`. I'll do this.

include/public/osqp_api_functions.h
imciner2: Yea, you are right, this change actually doesn't make sense considering I suggested to add the error check in the comment below this :laughing:. We can leave these returning ints and use it to do error handling.

include/public/osqp_api_functions.h
vineetbansal: Comments added

src/osqp_api.c
vineetbansal: Done

src/derivative.c
vineetbansal: Done

src/derivative.c
vineetbansal: Done

src/derivative.c
vineetbansal: Done

src/osqp_api.c
vineetbansal: Done

vineetbansal:
@imciner2 - I've made the changes. All tests still pass. There was a `OSQPVectorf_subvector_assign` that I'm using now (much better!), though I had to introduce a `multiplier` argument to it since that's the common use case here. Also added a `OSQPVectorf_subvector_assign_scalar` since that's another thing happening here.

Unrelated to this PR, but I noticed that the following code was failing to compile under certain conditions (when I was trying to compile `osqp` in debug mode), and gives a `else without if` or a similar error:

```
void print_dns_matrix(const OSQPFloat* M,
...
      if (j < n - 1)
        // c_print("% 14.12e,  ", M[j*m+i]);
        c_print("% .3g,  ", M[j * m + i]);

      else
        // c_print("% 14.12e;  ", M[j*m+i]);
        c_print("% .3g;  ", M[j * m + i]);
    }
```

I take it that `c_print` effectively disappears under certain conditions - adding parenthesis around the `if/else` works fine.

src/derivative.c
vineetbansal: done

vineetbansal:
Ah - looks like I need to `ifdef` some code for non-default algebras. I should be able to do this before our meeting tomorrow.

imciner2 (approved):
Lets go ahead and merge this and continue iterating in the branch.

vineetbansal:
Thanks @imciner2 - I see that my runner has again been deleted from the repo. Do you know how this might have happened again? If no one actively did this perhaps there's some auto monitor stuff that's doing this..

imciner2:
> I see that my runner has again been deleted from the repo. Do you know how this might have happened again? If no one actively did this perhaps there's some auto monitor stuff that's doing this..

It appears that GitHub will automatically remove a runner that hasn't connected within a certain time, their docs say (https://docs.github.com/en/actions/hosting-your-own-runners/removing-self-hosted-runners):

> A self-hosted runner is automatically removed from GitHub if it has not connected to GitHub Actions for more than 14 days.

So I'm guessing it disconnected over the break and then was removed.

PR #488, opened 2022-11-16
Simplify projection functions

imciner2:
After the introduction of the algebra API layer, the projection functions were essentially thin wrappers over vector API calls. This added an additional layer of complexity and indirection, so instead let's just remove that extra layer and use the vector API directly where needed (since each function was only called from one place in the code anyway, and they are basically single line trampoline functions).

This also has the advantage of removing a source file and header file, so there are two fewer files that have to be included with codegen/embedded projects now.


bstellato (approved):

bstellato:
LGTM! Thanks. Let's merge this after the runner completes the job successfully.

PR #490, opened 2022-11-17
Running CI on macos-11 + macos-latest

vineetbansal:
To see if its a macos-12 issue

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
         fail-fast: false
 
         matrix:
-          os: [ubuntu-latest, macos-latest, windows-latest]
+          os: [ubuntu-latest, macos-latest, macos-11, windows-latest]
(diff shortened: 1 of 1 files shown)


vineetbansal:
All tests on `macos-11` are working. It seems like its an XCode 14 error on the latest macos image, as per https://github.com/actions/runner-images/issues/6350.

We could try to use `macos-latest` and downgrade XCode as per:
https://github.com/actions/runner-images/issues/6350#issuecomment-1268843074

or alternatively, pin to `macos-11` for now?

vineetbansal:
I've pinned the CI to macos 11 for now. Perhaps we can open up an issue to move to macos-latest (the failures on which have nothing to do with our code AFAIK).

imciner2:
Is macOS 11+ support an issue on the develop-1.0 branch? I think this has to do with how we are linking the MKL library file, which changed in the upcoming version.

PR #491, opened 2022-11-17
Update main-cuda.yml

vineetbansal:
WIP - ignore till CUDA runner works on this PR again.

### DIFF
--- a/.github/workflows/main-cuda.yml
+++ b/.github/workflows/main-cuda.yml
@@ -39,13 +39,6 @@ jobs:
             lfs: false
             submodules: recursive
 
-        - name: Set up conda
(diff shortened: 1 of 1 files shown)


vineetbansal:
Ok - I think this can be squashed and merged. The change here is that I'm not installing `conda` on the runner from scratch, which may fail for reasons that have nothing to do with our code, as was the case recently with `libtinfo.so` errors (these were caused by outdated libtinfo/libstdc++ libraries being pulled from `conda` upstream). I think it's better to use a pre-configured conda environment which has weeded out these issues ahead of time.

imciner2 (approved):

PR #492, opened 2022-11-17
CUDA: Don't allow creating cusparse matrices with a 0 dimension

imciner2:
Some versions of cuSPARSE have problems with creating dense vectors with length 0 (they say it is an invalid input) and sparse matrices with any dimension of 0 length, so we need to avoid creating them. Instead, don't create them and just never perform the operations if there are no values in the matrices.

Hopefully this will fix the assertions we are seeing with the CUDA CI. (I can't test that locally because I haven't been able to trigger those on my machine, only on the CI machine).

### DIFF
--- a/algebra/cuda/matrix.cu
+++ b/algebra/cuda/matrix.cu
@@ -105,7 +105,10 @@ void OSQPMatrix_Axpy(const OSQPMatrix*  mat,
     cuda_vec_mult_sc(y->d_val, beta, y->length);
     return;
   }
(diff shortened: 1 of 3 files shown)


imciner2:
Ok, @vineetbansal this seems to fix the CUDA issue on your CI machine, and the test failure that is left is going to be fixed by https://github.com/osqp/osqp/actions/runs/3490448148/jobs/5841892325. So the combination of these two PRs will make the CUDA tests passing in CI now.

Issue #334, opened 2021-09-16
Add automatic link checker to osqp.org website


bstellato:
It should help with issues like #329 

This should be a good option: https://github.com/filiph/linkcheck

imciner2:
Here is an example of a link checker workflow that runs periodically: https://github.com/smartcontractkit/documentation/blob/222687e8acf6b51669724e2513ea51df32b64a19/.github/workflows/check-links.yml.

I think having a workflow like that (in addition to a build-time test) would be good so that we can catch when external links break due to external changes.

PR #508, opened 2023-03-29
Add workflow to automatically check website links

imciner2:
This adds a simple workflow that will run at 2am everyday to test all the links on the OSQP website (both internal links and external links) to ensure they are valid. It operates by crawling the hosted website instead of our local copy, so we can get in-situ results and it can check both the stable and develop-1.0 docs as well, hopefully.

I need to push this to the master branch directly to allow the scheduled workflow to run properly (scheduled workflows only operate on the main/default branch).

Fixes https://github.com/osqp/osqp/issues/334

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -1,0 +1,18 @@
+name: Documentation and Site Link Checker (Periodic)
+
+on:
+  schedule:
+    # Run daily at 2am
+    - cron: '0 2 * * *'
+
+jobs:
+  check-links-periodic:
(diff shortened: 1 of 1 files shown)


vineetbansal:
Looks good. I didn't know one can run periodic jobs on github actions. Good to know.
I think it makes sense to merge this once PR #490 is merged in. Unless we see all green check marks on these PRs, we'll never know if we've broken something.

imciner2:
@vineetbansal this is now passing all CI after #490 was merged, so if you can approve this we can get the pipeline running on the current site and see how it works.

PR #509, opened 2023-03-29
Bump version of checkout action to support node 16 by default

imciner2:
GitHub has deprecated actions built on node 12 for their shared runners, so it now triggers warnings on every pipeline they are used. actions/checkout supports node 16 on v3, so bump our version from v2 to v3.

These deprecation warnings can be seen in run https://github.com/osqp/osqp/actions/runs/4554774592.


imciner2:
And Windows MKL is broken on this one... why is it always the MKL libraries that give us trouble :disappointed:

PR #516, opened 2023-04-27
Using latest docs marketplace action

vineetbansal:
The `_sources` is being copied inside the `docs` folder correctly, as can be seen at:
https://github.com/osqp/osqp/tree/gh-pages/docs/_sources

I'm not sure why we're still getting a 404 on the `*.rst.txt` files. I'm upgrading to the latest version of the marketplace action here just for investigation.


vineetbansal:
Closing this to avoid interference with ongoing (unrelated) DNS issues.

PR #519, opened 2023-05-12
Website fixes

imciner2:
- Update a paper link
- Change the edit link on the top of the docs pages to go directly to GitHub instead of a different file (which is missing now anyway)
- Update contributors


vineetbansal (approved):
Thanks!

PR #522, opened 2023-05-16
Implement more MKL algebra vector functions using VML functions

imciner2:
Calling the Vector Math functions should provide more performance on some systems compared to hand-rolled loops. This isn't always true however, so we can't just replace everything. For instance, replacing a single for loop with two VML calls (such as for the elementwise bounding) might actually be more inefficient than the loop.

Also switch over to using the cblas wrapper to simplify the calling code (no need to pass pointers to integers, just pass the integers themselves).

### DIFF
--- a/algebra/mkl/blas_helpers.h
+++ b/algebra/mkl/blas_helpers.h
@@ -3,8 +3,9 @@
 
 #include "osqp_configure.h"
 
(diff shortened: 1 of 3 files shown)


vineetbansal:
python tests pass against this branch, so I think it can be merged.

PR #525, opened 2023-05-17
Don't change the external API for codegen and derivatives

imciner2:
To preserve compatibility across different library versions compiled with different algebra backends, don't actually remove the codegen and derivative parts from the public API when they aren't implemented. Instead, return an error code saying the functions aren't implemented.

This way, any program compiled with a non-embedded OSQP version can be used with a static/shared library version from any backend without having to recompile or make any code changes.


vineetbansal:
Thanks @imciner2 - python tests pass if I build them against this branch (and make the necessary changes), so I'm merging this.

PR #536, opened 2023-05-30
Merge 1.0 development into master to make first beta release

imciner2:
This PR is the final push in the 1.0 development cycle. When we merge this in, I will create the first beta release for 1.0 on the master branch.

PR #537, opened 2023-05-31
Merge 1.0 development into master to make first beta release

imciner2:
This PR is from a branch that has had master merged into it already, so now GitHub's web UI seems to accept it as valid for merging to master. This will replace PR https://github.com/osqp/osqp/pull/536, and be the merging of all 1.0 development into master.

We need to make sure all runners and jobs complete properly to ensure everything looks good before merging.


vineetbansal (approved):
Awesome @imciner2 . Let's do this!

PR #559, opened 2023-06-23
CMake cleanup and fixes

imciner2:
Cleanup various parts of the CMake build system and CMake installation components.

Changes include:
* Remove custom CMake script for locating Intel oneAPI and use the intel provided one (this requires manually setting the CMAKE_PREFIX_PATH variable to find MKL though due to a bug in Intel's environment scripts)
* Make sure dependencies of the static libraries are passed to the applications compiling them using the CMake configure scripts
* Use actual CMake targets for CUDA libraries so the dependencies are passed properly in the CMake configure scripts
* Ensure the static library is linked to the consuming application using C++ when using the CMake configure scripts (CUDA requires the C++ standard library, which comes from the C++ compiler)


vineetbansal:
Hi @imciner2 - I'm able to build wheels for all algebras against the `im/cmake` branch, and all the tests still pass, so it all looks good.

imciner2:
@vineetbansal great! I just merged this into the queue branch for the next beta release.

PR #563, opened 2023-07-13
Fix build warnings

imciner2:
Fix various warnings found when building with GCC `-Wall` flag.

Main changes are:
* Removing unused variables
* Modifying the format string used to write the integer based on the current `OSQPInt` type to make sure values are always representable

### DIFF
--- a/algebra/_common/kkt.c
+++ b/algebra/_common/kkt.c
@@ -303,11 +303,6 @@ OSQPCscMatrix* form_KKT(OSQPCscMatrix* P,
   OSQPInt   m,n;            //number of variables, constraints
   OSQPInt  nKKT, nnzKKT;    // Size, number of nonzeros in KKT
   OSQPInt  ndiagP;          // entries on diagonal of P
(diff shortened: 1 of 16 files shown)


govindchari:
There are still some warnings when 
`-Wextra` is enabled. The warnings are unused parameters and an implicit fallthrough.

imciner2:
The unused parameters are intentional. The way the algebra system is built, all implementations need to use the same function prototypes, so there will be some values used by backends such as CUDA that aren't needed in the builtin algebra backend, and some main features that aren't implemented in some backends (but the API functions are still present to allow for a consistent API when the algebra backend changes). I've added a macro that hides these warnings now.

The implicit fall through warning is just GCC being absolutely stupid. There is a comment there describing the fall through already, but GCC doesn't recognize that because it insists that any comment declaring a fall through must contain only some permutation of "fall through" (or other restrictive regexes), so it doesn't recognize the current comment because I actually explained why the fall through is needed. I've added a second comment that does meet the GCC regex, so that will quiet that warning now.

Denyweeeed (approved):

PR #564, opened 2023-07-14
Replaced incorrect and statement with or

govindchari:
There was an incorrect and statement that should have been an or in the check on conjugate gradient fraction.


imciner2 (approved):
Thanks for catching this. Looks like the original author thought the inequalities were the other way around.

PR #577, opened 2023-08-04
Improve public API for user-facing types

imciner2:
We have three main user-facing types that are used to pass data into the solver: `OSQPSettings`, `OSQPCscMatrix` and `OSQPCodegenDefines`. Currently, our system requires the user to malloc the object, then do a separate call to get the defaults, then use them. This can be very awkward, and leads to having a lot of boilerplate code like

```
OSQPCscMatrix* P = malloc(sizeof(OSQPCscMatrix));
OSQPCscMatrix* A = malloc(sizeof(OSQPCscMatrix));

/* Populate matrices */
csc_set_data(A, m, n, A_nnz, A_x, A_i, A_p);
csc_set_data(P, n, n, P_nnz, P_x, P_i, P_p);
```

or 
```
settings = (OSQPSettings *)malloc(sizeof(OSQPSettings));

osqp_set_default_settings(settings);
settings->polishing = 1;
```

With this PR, new functions to allocate and free these objects are added:
* `OSQPCscMatrix_new()`
* `OSQPCscMatrix_free()`
* `OSQPSettings_new()`
* `OSQPSettings_free()`
* `OSQPCodegenDefines_new()`
* `OSQPCodegenDefines_free()`

Additionally, the `csc_set_data()` function has been renamed to `OSQPCscMatrix_set_data()`.

These changes mean that the boilerplate code from above becomes:
```
/* Create CSC matrices that are backed by the above data arrays. */
OSQPCscMatrix* P = OSQPCscMatrix_new(n, n, P_nnz, P_x, P_i, P_p);
OSQPCscMatrix* A = OSQPCscMatrix_new(m, n, A_nnz, A_x, A_i, A_p);
```
and
```
OSQPSettings* settings = OSQPSettings_new();
settings->polishing = 1;
```

These 6 new functions are only available in non-embedded mode since they allocate data. The `OSQPCscMatrix_set_data()` function is still available in all modes, since it will just assign pointers (so it is still a direct replacement for `csc_set_data()`.


goulart-paul:
Agree that this is a bit nicer, but with this:

`/* Create CSC matrices that are backed by the above data arrays. */
OSQPCscMatrix* A = OSQPCscMatrix_new(m, n, A_nnz, A_x, A_i, A_p);`

It needs to be really clear what happens when I call `OSQPCscMatrix_free`.  Specifically, the call to the `_free` is (presumably?) only freeing the container and not any of the data arrays that were passed to `_new`.   Otherwise you get issues with cases like the examples that use stack allocated arrays.

It would also be nice to provide something like `OSQPCscMatrix_identity(n)` or `OSQPCdcMatrix_zeros(m,n)`, but then the behaviour of the `_free` is confusing because the user would be required to free internal arrays that they hadn't directly allocated.

We could maybe add some internal data ownership flag to the `CscMatrix` type that indicates whether it's `free` function should also free its arrays.

imciner2:
I've gone ahead and tweaked the API so that we now track owning and not-owning the arrays inside a CSC matrix, and free them accordingly. This comes alongside adding some helper functions to create diagonal matrices, the identity matrix and a matrix of all zeros.

bstellato (approved):

bstellato:
WIth the new changes it looks great to me! @goulart-paul what do you think? I saw you left quite a few comments

Issue #581, opened 2023-08-31
Implement profiling annotations


imciner2:
Both VTune and NVidia's Nsight have the ability to add annotations to blocks of code showing what they are doing. These annotations are then shown in the timeline and allow for easier viewing of the stages of the program. A similar system to the algebra one should be built for the profile annotations.

For VTune:
* Uses the ITT API
* Build docs: https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2023-0/configuring-your-build-system.html
* API reference: https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2023-0/instrumentation-tracing-technology-api-reference.html
* Compile library is included with VTune

For Nsight:
* C API reference: https://nvidia.github.io/NVTX/doxygen/index.html
* Able to be found using CMake and the `CUDA::nvtx3` target (https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cuda-toolkit-nvtx3) - CMake 3.25+
* Old example showing usage: https://developer.nvidia.com/blog/cuda-pro-tip-generate-custom-application-profile-timelines-nvtx/

imciner2:
Support for profiler annotations has been merged into the development branch now.

PR #585, opened 2023-09-27
Introduce profiler annotations

imciner2:
Some profilers (like NVidia Nsight and Intel VTune) provide the ability to add labels to sections of the code to track how long the program is spending in each part of the code. This can then be used to introduce labels on the flamegraph or generate statistics about the timing/percentages.

This implements a system that allows for compile-time removal of these (if no library is available), and also for choosing how detailed the annotations should be at compile time. There are three levels of detail:
* 0 = No annotations
* 1 = OSQP ADMM only
* 2 = OSQP ADMM and linear algebra linsys solver

For the CUDA backend, this uses the NVidia-provided NVTX library, and can displays results in Nsight. For instance, a sample result using this with the OSQP demo program is:

image

Fixes https://github.com/osqp/osqp/issues/581


imciner2:
Added annotations for the Intel VTune compiler (through ITT). Unfortunately, that doesn't help with AMD processors (VTune doesn't play nice with non-Intel processors).

There is annotation support for the AMD ROCm profiler though, [roctracer](https://github.com/rocm-developer-tools/roctracer). The ROC-TX library in that repo contains similar push/pop calls to add annotations.

imciner2:
Now this has support for Intel VTune, AMD OmniTrace, and NVidia Nsight profiling annotations. I have included the code for the rocprofiler as well (using the ROCTX library), however that actually only supports AMD GPUs, so it isn't much use to us right now (but might be in the future, so I left it in).

This also now has a new documentation section for "Advanced Features" to document all these profiling tools and the integration. I am hoping to be able to also use some of these with the Julia wrapper library, so in the future the docs might be expanded to also include instructions for that once I figure out how to do it. For now, this should be complete though.

Issue #246, opened 2020-04-12
Add Java as target


davidpham87:
Hello,

I wondered if it would make a lot of work to add Java as a target for the C code? I am more interested in Clojure, but exporting to Java would allow all JVM language (and Android), to use your library.

Regards,
David

goulart-paul:
It is certainly possible, and the reason it has not been done is simply lack of developer bandwidth.   We are open to someone writing such an interface for the project though, as happened with the R and Rust interfaces.

loehndorf:
We did a port of OSQP to pure Java. Performs at par to the C lib by our own measurements. Here's the link: https://github.com/quantego/josqp

PR #595, opened 2023-11-23
Add list of community-developed implementations of our ADMM algorithm

imciner2:
There are now a few community-developed implementations of the ADMM algorithm used by OSQP, so lets add a small table on the algorithm description page listing them and providing links to them.

What is written:
image

cc @loehndorf, this adds the jOSQP implementation into our documentation website.

### DIFF
--- a/docs/interfaces/index.rst
+++ b/docs/interfaces/index.rst
@@ -1,3 +1,5 @@
+.. _interfaces:
+
 Interfaces
 ============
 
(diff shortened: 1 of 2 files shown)


bstellato:
This looks great! This should (partially) solve https://github.com/osqp/osqp/issues/246  As long as it appears on the left column as "alternative implementations", I think this is ready to go!

bstellato (approved):

PR #605, opened 2024-02-20
Consolidate dy_l and dy_u into a single dy for adjoint derivatives, and other fixes

imciner2:
The need for both lower and upper bound perturbations of the dual variables is an implementation detail of the current implicit differentiation method, and does not always need to be the case. Instead, the user-facing API should just have a single dy input and the differentiation method should then handle the duplication internally.

Also, this adds some basic smoke tests to ensure the C API compiles and does not crash when performing the derivative operations. There is no numerical testing of the computations yet though.

An additional part of this was to fix the basic_qp tests when using ASAN, since they had some data lifetime issues. Now the entire test suite can run under ASAN with no problems reported.


imciner2:
Test failure is in numpy for some reason, but all other tests pass. That one has been flaky in the past though.

PR #606, opened 2024-02-20
Enable dependabot on GitHub actions workflows

imciner2:
Lets try this to keep on top of the GitHub actions versions, since we are having some deprecation warnings for actions in pipelines right now.


vineetbansal (approved):
Thanks!

PR #618, opened 2024-03-25
Manage Memory allocation errors in QDLDL

### LINKED ISSUE #587: Unhandled allocation errors in qdldl_interface.c
Hi,

We are facing some unhandled memory allocation errors in `qdldl_interface.c`.
For example, in the method `init_linsys_solver_qdldl`, the qdldl solver is allocated and used this way:

```CPP
    // Allocate private structure to store KKT factorization
    qdldl_solver* s = c_calloc(1, sizeof(qdldl_solver));
    *sp = s;

    // Size of KKT
    n = P->csc->n;
    m = A->csc->m;
    s->n = n;
    s->m = m;
```

But what about `s->n` and `s->m` if the allocation fails and `s` is null?

Such allocation operations are checked in other parts of the code, like for example in `osqp_api.c`, where we can find such statements (method `osqp_setup`):

```CPP
  // Allocate empty solver
  solver = c_calloc(1, sizeof(OSQPSolver));
  if (!(solver)) return osqp_error(OSQP_MEM_ALLOC_ERROR);
  *solverp = solver;
```

Is there a plan to add such sanity check in `qdldl_interface.c` in the future?
For example, we can see the `init_linsys_solver_qdldl` returns `OSQP_NONCVX_ERROR` in case the problem is not convex.
Could it be possible to return `OSQP_MEM_ALLOC_ERROR` as soon as an allocation operation fails?

On our side, it would prevent a ugly crash and allow us to manage the situation in a smoothier way.

Thanks a lot in advance.

david-g-sap:
Each time memory is allocated, the returned pointer is checked.
If it is null, the method ```init_linsys_solver_qdldl``` returns the dedicated error code ```OSQP_MEM_ALLOC_ERROR```

This Pull Request fixes #587

### DIFF
--- a/algebra/_common/lin_sys/qdldl/qdldl_interface.c
+++ b/algebra/_common/lin_sys/qdldl/qdldl_interface.c
@@ -99,6 +99,12 @@ static OSQPInt LDL_factor(OSQPCscMatrix* A,
     // Allocate memory for Li and Lx
     p->L->i = (OSQPInt *)c_malloc(sizeof(OSQPInt)*sum_Lnz);
     p->L->x = (OSQPFloat *)c_malloc(sizeof(OSQPFloat)*sum_Lnz);
+    if (!(p->L->i) || !(p->L->x))
+    {
+        // Don't free p->L->i / p->L->x here, as they will be released when calling
+        // csc_spfree(s->L) in free_linsys_solver_qdldl
+        return OSQP_MEM_ALLOC_ERROR;
+    }
     p->L->nzmax = sum_Lnz;
 
     // Factor matrix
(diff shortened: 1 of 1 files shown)


imciner2:
I believe this should already be fixed by https://github.com/osqp/osqp/pull/573, which is in the develop-beta1 branch currently. (that branch contains the development after the most recent beta release).

david-g-sap:
> I believe this should already be fixed by #573, which is in the develop-beta1 branch currently. (that branch contains the development after the most recent beta release).

Thanks a lot for the information.
I will close this PR and try to test the version from develop-beta1

imciner2:
Note that I just merged develop-beta1 into master, so you should see these memory allocation changes in the master branch now.

david-g-sap:
> Note that I just merged develop-beta1 into master, so you should see these memory allocation changes in the master branch now.

Thanks a lot for the information.

Is the release of version 1.0.0 already planned?

david-g-sap:
Sorry, but it seems the PR #573 does not include any changes in the method ```init_linsys_solver_qdldl()```.
I can still see dynamically allocated pointers that are used without being tested.
In my PR proposal, I made changes in ```init_linsys_solver_qdldl```, ```LDL_factor``` and ```permute_KKT```.
But mainly in ```init_linsys_solver_qdldl``` to prevent core dump after allocation errors.

PR #632, opened 2024-04-04
Add dual objective and duality gap computation

### LINKED ISSUE #175: Return dual objective function
Find an easdy way to return a dual objective function value without having to compute the inverse of `P`. This is related to the discussion in https://github.com/oxfordcontrol/OSQP.jl/pull/55

Suggestions are welcome!

imciner2:
This adds computation of the dual objective value and the duality gap (distance between the primal and dual objective values). Note that we don't actually do `primal_val - dual_val` for the duality gap computation because near optimality it could lead to numerical issues (since the two values will be very close together, catastrophic cancellation could occur). Instead, we compute the dual objective value as `-0.5*x^T P x - SC(y)` and the duality gap as `x^T P x + q^T x + SC(y)`. Additionally, the integral of the duality gap over the runtime of the solver is reported. (this value could be a metric that is used to decide how quickly the algorithm is converging).

This also adds a termination criterion for duality gap value, so now the optimal point must have three values within tolerance: primal residual, dual residual and duality gap.

This is reported inside the `OSQPInfo` struct, and also in the output displayed to the user at the end of the solve.

Sample output:
```
-----------------------------------------------------------------
           OSQP v1.0.0.beta0  -  Operator Splitting QP Solver
              (c) Bartolomeo Stellato,  Goran Banjac
        University of Oxford  -  Stanford University 2021
-----------------------------------------------------------------
problem:  variables n = 2, constraints m = 3
          nnz(P) + nnz(A) = 7
settings: algebra = Built-in,
          linear system solver = QDLDL v0.1.6,
          eps_abs = 1.0e-03, eps_rel = 1.0e-03,
          eps_prim_inf = 1.0e-04, eps_dual_inf = 1.0e-04,
          rho = 1.00e-01 (adaptive),
          sigma = 1.00e-06, alpha = 1.60, max_iter = 4000
          check_termination: on (interval 25),
          time_limit: 1.00e+10 sec,
          scaling: on, scaled_termination: off
          warm starting: on, polishing: on,
iter   objective    prim res   dual res   gap        rho        time
   1  -1.5499e-02   1.02e+00   1.01e+02  -1.02e+02   1.00e-01   5.47e-04s
  75   1.8798e+00   1.12e-03   2.80e-06  -2.22e-04   1.00e-01   6.22e-04s
plsh   1.8800e+00   0.00e+00   3.14e-16  -4.44e-16   --------   8.59e-04s

status:               solved
solution polishing:   successful
number of iterations: 75
optimal objective:    1.8800
dual objective:       1.8800
duality gap:          -4.4409e-16
primal-dual integral: 1.5466e-03
run time:             8.59e-04s
optimal rho estimate: 3.91e+00
```

Fixes #175


imciner2:
@bstellato this looks like it should be good to go since all tests are passing now. If we merge this in, I can make a new beta2 release of the C library shortly after.

PR #663, opened 2024-08-13
Fix the CI involving MKL

imciner2:
The Intel conda repos changed locations, so now we must install using the full Intel URL instead of a specific anaconda channel. Additionally, we only need MKL when doing MKL tests in coverage, it is extraneous for the other backends and doesn't need to be installed.


vineetbansal:
@imciner2 - thanks for investigating and fixing this - these `conda` steps will also come in handy for the documentation on building the wrappers.

vineetbansal (approved):

Issue #646, opened 2024-05-15
Codegen: updating all entries of P or A


maxschaller:
In version 0.6, updating all entries of the data matrices was accomplished by calling

`osqp_update_P(&workspace, values_P, 0, 0);` and `osqp_update_A(&workspace, values_A, 0, 0);`

where `values_P` and `values_A` are the new `csc` data vectors and `0` is passed as the last argument (number of values to update) to imply that all nonzeros should be updated.

In version 1.0.0b1, this is achieved via

`osqp_update_data_mat(&solver, values_P, 0, nnz_P, values_A, 0, nnz_A);`

where `nnz_P` and `nnz_A` are the respective numbers of nonzeros in `P` and `A`. If `0` is passed like in version 0.6, nothing will be updated.

This issue asks to change the behavior back, such that `0` as the fourth and last argument implies that all nonzeros are updated.

PR #666, opened 2024-08-13
Re-add support for passing nnz=0 when updating all matrix values

imciner2:
This was allowed in 0.6, and it makes sense as a behavior, so re-add support for it. Except handle this in the main OSQP function instead of algebras, that way algebras only need to handle one update case.

CC @maxschaller 

Fixes https://github.com/osqp/osqp/issues/646


imciner2:
MKL is failing for an unrelated reason... it is being very weird. Merging this because the test failure won't have been affected by it.

PR #667, opened 2024-08-14
Refactor how rho update works

imciner2:
This refactors the adaptive rho settings and code to make it cleaner in the code, and improve the settings interface for the user.

This lets the user choose a specific method for updating rho specifically using the `adaptive_rho` variable instead of having it be a combination of other settings. This allows choosing either fixed iteration interval updates or the time-based update. Additionally, this prints out the summary line for the iteration where a rho update occurred to make it visible this happened.

Note, this is based on PR #632 and contains that code as well.

### DIFF
--- a/src/osqp_api.c
+++ b/src/osqp_api.c
@@ -697,7 +697,7 @@ osqp_profiler_sec_push(OSQP_PROFILER_SEC_OPT_SOLVE);
     can_adapt_rho = 0;
 #endif /* OSQP_EMBEDDED_MODE != 1 */
 
-    if(can_check_termination || can_print || can_adapt_rho || iter == 1 || settings->restart_enable) {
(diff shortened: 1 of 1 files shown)


examples/osqp_demo.c
```diff
@@ -60,16 +178,21 @@ int main(void) {
...
-  if (P) free(P);
-  if (settings) free(settings);
+
+  if( dynamic_matrices ) {
+    if (qp_A) free(qp_A);
+    if (qp_P) free(qp_P);
+  }
+
+  if (settings && dynamic_settings)
+    free(settings);
```
AmitSolomonPrinceton, line 195: Didn't we have c_free for our on data structures?
(This comment applies to other places in the code where we use free)

PR #693, opened 2024-12-05
Fixed broken link in documentation

govindchari:
The existing link to the file where constants were defined was broken. Previously it pointed to `include/constants.h` but it should be pointing to `include/public/osqp_api_constants.h`.


imciner2 (approved):

imciner2:
Thanks!

PR #699, opened 2025-01-11
Fixed hyperlink on home page and python build instructions

govindchari:
Previously the python build instructions did not work. I assume the build process for the python wrapper changed at some point, but the documentation was not updated.


vineetbansal:
@govindchari - thanks!

PR #700, opened 2025-01-16
docs: update setting interfaces name for examples

bxtbold:
I cross-checked and found that the naming for [`warm_starting`](https://osqp.org/docs/interfaces/solver_settings.html) is different in other interfaces, including Python, Julia, and MATLAB. Instead, it is referred to `warm_start`

I have updated the docs examples in this PR to incorporate this issue.

### DIFF
--- a/docs/examples/lasso.rst
+++ b/docs/examples/lasso.rst
@@ -64,7 +64,7 @@ Python
     prob = osqp.OSQP()
 
     # Setup workspace
-    prob.setup(P, q, A, l, u, warm_starting=True)
+    prob.setup(P, q, A, l, u, warm_start=True)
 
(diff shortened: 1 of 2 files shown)


bxtbold:
@vineetbansal thank you for your review. I agree with your opinion to close the PR.

I have installed [osqp-python](https://github.com/osqp/osqp-python) from the source `warm_starting` seems to be working okay. However, it is not the case when I pip installed it. It seems like the internal low level osqp module has some bug. Maybe I will open an issue or PR in osqp-python if there is something I can do with it.

Error:
```bash
self._model.setup(*unpacked_data, **settings)
TypeError: 'warm_starting' is an invalid keyword argument for this function
```

Since this is nothing related to this repo, I am closing it.

PR #709, opened 2025-03-07
fixed links reported by link checker

vineetbansal:
I've been getting daily emails from our link checker CI about broken links for a couple of years now :). This PR fixes these.

There's also an issue with images not being displayed correctly (for example `Sample VTune timeline panel.png` in https://osqp.org/docs/advanced/profiling.html. The logic is just fine when I run this locally, so I'm hoping that adding the `.nojeyll` file at the root of the `gh-pages` branch (not at the `site/` folder which we ask the gh action to deploy) fixes the issue.

The [Interfacing new system solver](https://osqp.org/docs/contributing/solvers.html) page is very likely outdated, but that's a separate issue for which I'll open another issue.


imciner2:
You get emails about this. I never got any, and only noticed these failures a few weeks ago but didn't have the chance to go through and fix them, so thanks for doing this. I'll review it later today hopefully.

docs/advanced/profiling.rst
```diff
@@ -142,7 +142,7 @@ to report sections of the code for timing. To enable the section reporting, use
 CMake option during configuration.
 
 
-.. _AMD OmniTrace: https://amdresearch.github.io/omnitrace/index.html
+.. _AMD OmniTrace: https://rocm.docs.amd.com/projects/rocprofiler-systems/en/latest/
```
imciner2, line 145: Yep, AMD renamed all their tooling related to the profiling systems, so I actually don't know how accurate any of this AMD stuff is as of right now. When I spoke to some people at SIAM CSE last week, it sounded like they actually collapsed the omnitrace API here into rocTX for the rocprofiler-system, so it might be this separate part isn't needed anymore.

docs/citing/index.rst
```diff
@@ -43,7 +43,7 @@ If you use OSQP for published work, we encourage you to put a star on `GitHub <h
           }
 
     GPU implementation
-        GPU implementation and PCG method for solving linear systems in this `paper <https://doi.org/10.1016/j.jpdc.2020.05.021>`__.
+        GPU implementation and PCG method for solving linear systems in this `paper <https://www.sciencedirect.com/science/article/pii/S0743731520303063>`__.
```
imciner2, line 46: I'd really prefer to keep the DOIs if at all possible. Do we know why the link checker is flagging the DOI-based links? Can it not handle the fact they are redirects?

docs/contributing/python.rst
```diff
@@ -82,7 +82,7 @@ The process of building the ``MKL`` OSQP wrappers is the same as above, except t
 Wrappers for ``cuda`` algebra
 --------------------------
 
-For building Python wrappers for the ``CUDA`` algebra backend, you need to have the `CUDA Tooklit<https://developer.nvidia.com/cuda-toolkit-archive>`_.
+For building Python wrappers for the ``CUDA`` algebra backend, you need to have the `CUDA Tooklit <https://developer.nvidia.com/cuda-toolkit-archive>`_.
```
imciner2, line 85:
```suggestion
For building Python wrappers for the ``CUDA`` algebra backend, you need to have the `CUDA Toolkit <https://developer.nvidia.com/cuda-toolkit-archive>`_.
```

docs/get_started/linear_system_solvers.rst
```diff
@@ -34,10 +34,10 @@ QDLDL is a sparse direct solver that works well for most small to medium sized p
 However, it becomes not really efficient for large scale problems since it is not multi-threaded.
 
 
-MKL Pardiso
+oneMKL Pardiso
 -----------
-`MKL Pardiso <https://software.intel.com/en-us/mkl-developer-reference-fortran-intel-mkl-pardiso-parallel-direct-sparse-solver-interface>`_ is an efficient multi-threaded linear system solver that works well for large scale problems part of the Intel Math Kernel Library.
-Intel offers `free lincenses <https://software.intel.com/en-us/articles/free-mkl>`_ for MKL for most non-commercial applications.
+`oneMKL Pardiso <https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/onemkl-pardiso-parallel-direct-sparse-solver-iface.html>`_ is an efficient multi-threaded linear system solver that works well for large scale problems part of the Intel Math Kernel Library.
+Intel offers `free licenses <https://www.intel.com/content/www/us/en/developer/articles/tool/onemkl-license-faq.html>`_ for oneMKL for most non-commercial applications.
```
imciner2, line 40: We probably don't need to say this anymore since their licensing has basically become free (this sentence is from the original docs when it was a bit more restricted I believe).

imciner2 (commented):
Thanks, looks good overall, although my main concern is with the DOI links.

docs/citing/index.rst
vineetbansal: I agree - its just that the redirects were being reported by the link-checker, and I didn't pay close attention. I agree - the doi links are more canonical and what we should preserve anyway.

PR #711, opened 2025-03-12
debugging MKL failure on Windows 2022

vineetbansal:
just to trigger CI - do not merge


vineetbansal:
Not explicitly specifying `conda-forge` as the channel in the test environment (which causes all subsequent steps to use `conda-forge` as a viable channel anytime a `conda install` is encountered.

Since `conda-forge` was installing a bootstrap package on linux for glibc (`libstdcxx-ng`) causing things to "just work", we're doing this explicitly for linux now.

I can't say this is ideal, and I can't explain the recent error in `large_qp` in other branches, except to say that the mix-n-match approach between channels that was being used for mkl-devel and all of its associated dependencies may have been causing the `large_qp` error (which could have happened any other time but just happened to surface now). 

Nevertheless, the CI with these new changes does make sense in its own right, and can/should be merged so we can move forward with the rest of the closeout steps.

imciner2 (approved):

imciner2:
Thanks for tracking this down! My theory was it was something in MKL that was only being exercised in the `large_qp` test that was incompatible in that older version of MKL that was getting pulled into the evironment.

I do notice that now MKL 2023.x.0 is being installed though, so somehow we are stuck on that version.

PR #714, opened 2025-03-12
updated examples (except Matlab)

vineetbansal:
Fixing all examples in documentation

### DIFF
--- a/docs/examples/huber.rst
+++ b/docs/examples/huber.rst
@@ -41,7 +41,7 @@ Python
     from scipy import sparse
 
     # Generate problem data
-    sp.random.seed(1)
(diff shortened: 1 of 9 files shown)


vineetbansal:
With these tweaks to examples, and PR [59](https://github.com/osqp/osqp-matlab/pull/59) to `osqp-matlab`, I've verified that all these work in C/Python/Julia/R/Matlab/Yalmip.

imciner2:
Wow, Yalmip also works with no changes? That is great - I was worried we would have to get that updated.

imciner2 (approved):
Looks good. I'll make an issue to track figuring out a way to automatically test these examples.

PR #716, opened 2025-03-13
minimal change to compile to wasm32

vineetbansal:
With this tweak, I'm able to compile osqp to wasm32, and can subsequently add the wasm target for the `osqp-python` CI.


CMakeLists.txt
```diff
@@ -282,7 +282,11 @@ if(NOT MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0 -g")
     set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -G -g")
   else()
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3")
+    if (DEFINED EMSCRIPTEN)
+      # for some reason -O3 is causing functions to not be exported in Emscripten
+    else()
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3")
+    endif()
```
imciner2, line 289:
```suggestion
    if (NOT DEFINED EMSCRIPTEN)
      # for some reason -O3 is causing functions to not be exported in Emscripten
      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3")
    endif()
```

This should be equivalent and doesn't leave an empty control body.

imciner2 (commented):
Is there a simple CI check that could be added to build with WASM to make sure nothing regresses in the future?

imciner2 (approved):

PR #718, opened 2025-03-14
read all pages and updated docs

vineetbansal:
Other than the matlab codegen functionality that I haven't tested out (but something that @AmitSolomonPrinceton can vouch for), the documentation here is correct.

Added writeup on building/using backend algebras in python (currently in the wiki for this repo).

Added basic code examples for Rust (verified).


vineetbansal:
@imciner2 - I've added the migration guide and updated the matlab installation (from source) instructions, so I'm done with this PR for now.

The osqp-matlab installation instructions won't actually work for v1.0 (they'll install osqp 0.6), but will start working once we make a release on osqp-matlab.

PR #721, opened 2025-03-21
Bump to version 1.0.0

imciner2:
:tada: Version 1.0.0 release :tada:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,8 @@ jobs:
       fail-fast: false
 
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        # (macOS 13 is the last release to support x86 processors)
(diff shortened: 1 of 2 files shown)


vineetbansal:
Thanks @imciner2 - I see that the release workflow (which gets triggered on pushing tags and generates a bunch of downloadable files) hasn't been run since May 2024, when we pushed tag `v1.0.0.beta1`). It might need a couple of iterations in case something is broken on the CI, so just to be sure, I'll push tag `v1.0.0.beta2`.. etc till I get it working, after which I'll push tag `v1.0.0`

imciner2:
Sure, let's do a beta2 to see how it works. Hopefully that works and no new commits are needed.

vineetbansal (approved):

vineetbansal:
Fixed release workflow after testing on another branch. Added macos-13 to the matrix.

PR #727, opened 2025-03-31
Added bit about raise_errors in for solve method

vineetbansal:
Added a line about the optional parameter `raise_errors` as it currently exists in the python wrapper.


vineetbansal:
I've added instructions for using/switching/building the different algebra backends for the python wrappers in this PR. I won't be making any more changes to this PR (barring suggestions/tweaks), but will open up a new PR for the backends compatibility matrix documentation.

PR #729, opened 2025-04-02
Fix incorrect function name in v0.6.x to v1.0.0 migration guide

Jaanilj:
The migration guide incorrectly references `osqp_update_vec` as the replacement for `osqp_update_lin_cost` and `osqp_update_bounds` functions.

While I was migrating my own code, I noticed the correct API function name is `osqp_update_data_vec` as defined in `include/public/osqp_api_functions.h`.

This commit updates the documentation to reflect the correct function name.


imciner2:
Good catch!

PR #732, opened 2025-04-07
sample marimo notebook

vineetbansal:


site/wasm/osqp-1.0.0b4-cp312-cp312-pyodide_2024_0_wasm32.whl
imciner2, line 1: Since OSQP is upstream in pyodide, do we need to bundle a wheel in this repo? It seems like something that will be forgotten about and not updated regularly.

vineetbansal: oh yes that's why this is in draft mode - I need to investigate the relationship between marimo version and pyodide version.

vineetbansal:
@imciner2, @bstellato  - `marimo` is a few days behind and should pick up `pyodide==0.27.5` soon, but in any case for now I've generated the static webpages using a custom local build of `marimo` (everything in the `wasm/` folder here) as a starting point. The `.whl` file is gone and gets installed from `pyodide` automatically on import.

The generated marimo notebook is at:
https://osqp.org/wasm

A couple of thoughts:
- I generated this html folder locally but later we can integrate something in the CI (once marimo picks up the latest pyodide version automatically).
- We should make a more visually exciting version of the notebook but keep something like this basic example intact, something that helps one start using `osqp` immediately without getting distracted by domain-specific stuff.
- We can submit a PR towards adding `osqp` as a dependency for `cvxpy` wasm module later, but its still useful to have a plain vanilla notebook, because of the reason above.

.github/workflows/docs.yml
```diff
@@ -2,7 +2,7 @@ name: Docs
 
 on:
   push:
-    branches: [ master, release-0.6.3 ]
+    branches: [ master, release-0.6.3, vb/marimo ]
```
vineetbansal, line 5: This addition is only to build the `wasm/` folder using this branch, and should go away before we do an actual merge.

.github/workflows/docs.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: master
+          ref: vb/marimo
```
vineetbansal, line 20: This change is only to build the `wasm/` folder using this branch, and should go away before we do an actual merge.

bstellato:
This is great, especially to get started with the docs. That would help for sure new users.

@akshayka @mscolnick do you think we could add a more visual example in the marimo gallery?

akshayka:
We'd definitely welcome an OSQP notebook to our gallery — just let us know when you have something you'd like to share!

bstellato:
idea from @imciner2: it would be great to generate the static website from github actions so that we do not have to commit all the files.

bstellato (approved):
Great! Thanks for the great work @vineetbansal

PR #784, opened 2026-05-30
Guard "Polishing not needed" message behind settings->verbose

dug-jamieb:
The `c_print` in `polish()` for the no-active-set case is unguarded, unlike every other diagnostic print in `polish.c`.

As a result it prints on **every** solve that reaches an optimum with an empty active set, even when `verbose = 0`. Applications that solve repeatedly with polishing enabled get continuous stdout spam (one line per solve).

This wraps the message in the same guard used elsewhere in the file. No behaviour change beyond suppressing the message when `verbose` is off.

```c
  } else if (work->pol->n_active == 0) {
    /* No active constraints, so skip polishing */
#ifdef OSQP_ENABLE_PRINTING
    if (settings->verbose)
      c_print("Polishing not needed - no active set detected at optimal point\n");
#endif /* ifdef OSQP_ENABLE_PRINTING */
    info->status_polish = OSQP_POLISH_NO_ACTIVE_SET_FOUND;
```


dug-jamieb:
Hi, any reason for closing? Should I put up as a bug instead? Thanks

vineetbansal:
Hi @dug-jamieb - yes, please open an issue so we can judge the severity of the issue and go back and forth on it. The description above and the PR seemed to me to have been opened by AI, which makes it easily dismissable (I'm also going by your github history).

dug-jamieb:
Thanks for the response Vineet. It was an AI assisted PR but I revised it and cleaned it up. I looked but didn't see if you guys had an AI policy, apologies. I'll put it up as any issue, thanks :)

